PR #6, opened 2017-05-05
Remove occurences of old github org

bkabrda:


Jenkinsfile
```diff
@@ -18,7 +18,7 @@ node('docker') {
         dockerCleanup()
         // hack
         dir('worker') {
```
samuzzal-choudhury: @bkabrda are we having a directory worker in the server codebase to clone fabric8-analytics-worker?

msrb: This was necessary when the repos were private as docker couldn't clone them during build. This quick fix will enable CI/CD again. We can drop the worker hack later, it won't do any harm now :)

msrb:
[test this]

msrb:
@samuzzal-choudhury and we are green 😉

PR #17, opened 2017-06-01
Do not pin cucoslib commit

fridex:


fridex:
@msrb PTAL, I don't know why this variable was introduced. Is there some intention behind it?

msrb:
Actually yes, this was intentional. This way we can control what version exactly goes into the container. If you update cucoslib and you want to propagate the changes to api/jobs images, you'd need to do some completely unrelated change to the code to trigger CI/CD. Now you can just change the variable and it will be obvious what's going on. This clarity should make it easier to discover when the cucoslib update breaks api/jobs because people need to explicitly update the reference. 

And there is also docker caching. The step will be cached by docker so saying just "master" means people will be stuck with some old version of cucoslib, until they realize what is going on.

fridex:
OK, sounds reasonable. Let's explicitly do this kind of versioning.

Closing.

PR #31, opened 2017-06-16
Init Selinon when using configuration

fridex:


fridex:
@jpopelka PTAL, this should be easy

PR #42, opened 2017-06-23
Stop using usersys registry since it's EOL

pkajaba:


msrb:
Ah, you've just updated the PR :) cool, thanks 😉

PR #57, opened 2017-07-03
Multipom support

jyasveer:


bayesian/api_v1.py
```diff
@@ -308,7 +308,7 @@ def get(external_request_id):
             "started_at": started_at,
             "finished_at": finished_at,
             "request_id": external_request_id,
-            "result": manifest_response,
+            "stack_data": manifest_response,
```
samuzzal-choudhury: We should not change the schema ATM. Please revert to using "result".

PR #58, opened 2017-07-05
Adding the support to pick up manifest files from github url of repositories

GeetikaBatra:


GeetikaBatra:
As there are future plans to integrate multiple manifest files, there is a need of a mechanism to save the manifest files by unique names. But, currently manifest_descriptor checks if the name manifest name exactly matches with the type of manifest file e.g "pom.xml", "package.json" and "requirements.txt". In case of multiple manifest files, this name will conflict and it would be a tedious task to handle all the manifest files. One solution is to save the manifest files with a prefix of uuid. Another solution might be using hash for the file name. This would also involve improvising the manifest_descriptor validation. I am open to suggestions.

jpopelka:
Closing since it's been superseded by #115

PR #59, opened 2017-07-06
Maintain common bookkeeping for component and stack analysis

samuzzal-choudhury:


bayesian/utils.py
```diff
@@ -42,6 +54,10 @@ def server_create_analysis(ecosystem, package, version, api_flow=True, force=Fal
     :param force_graph_sync: force synchronization to graph
     :return: dispatcher ID handling flow
     """
+    # Bookkeeping first
+    component = MavenCoordinates.normalize_str(package) if ecosystem == 'maven' else package
+    server_create_component_bookkeeping(ecosystem, component, version, user_profile)
+
     args = {
         'ecosystem': ecosystem,
         'name': MavenCoordinates.normalize_str(package) if ecosystem == 'maven' else package,
```
fridex: This is already done above.

bayesian/utils.py
```diff
@@ -31,8 +32,19 @@ def server_run_flow(flow_name, flow_args):
     Setup.connect_if_not_connected()
     return run_flow(flow_name, flow_args)
 
+def server_create_component_bookkeeping(ecosystem, name, version, user_profile):
+    args = {
+        'external_request_id': uuid.uuid4().hex,
+        'data': {
+            'api_name': 'component_analyses',
```
fridex: Will it be run only on `component_analyses`?

bayesian/auth.py
```diff
@@ -15,6 +15,23 @@
 
 jwt.register_algorithm('RS256', RSAAlgorithm(RSAAlgorithm.SHA256))
 
+def decode_token():
+    token = request.headers.get('Authorization')
+    print (token)
```
samuzzal-choudhury: Removed. Changes are to be available in the next commit.

bayesian/auth.py
```diff
@@ -15,6 +15,23 @@
 
 jwt.register_algorithm('RS256', RSAAlgorithm(RSAAlgorithm.SHA256))
 
+def decode_token():
```
samuzzal-choudhury: @fridex need more inputs perhaps. AFAIK, decode_token now includes the entire decoding token logic and the login_required wrapper does not duplicate any of that. As you can see, login_required uses decode_token.

The current design of the decode_token allows it to be used in other places like stack-analyses and component-analyses logic in api_v1.py.

bayesian/auth.py
```diff
@@ -15,6 +15,23 @@
...
+    token = request.headers.get('Authorization')
+    print (token)
+    if token == None:
```
samuzzal-choudhury: Done. Changes are to be available in the next commit.

bayesian/api_v1.py
```diff
@@ -251,24 +251,29 @@ class ComponentAnalyses(ResourceWithSchema):
...
+            
         if ecosystem == 'maven':
             package = MavenCoordinates.normalize_str(package)
         package = case_sensitivity_transform(ecosystem, package)
         result = get_analyses_from_graph(ecosystem, package, version)
-        current_app.logger.warn("%r" % result)
 
         if result is not None:
             # Known component for Bayesian
+            server_create_component_bookkeeping (ecosystem, package, version, decoded)
```
samuzzal-choudhury: Done. Changes are to be available in the next commit.

bayesian/api_v1.py
```diff
@@ -472,6 +485,11 @@ class StackAnalysesV2(ResourceWithSchema):
 
     @staticmethod
     def post():
+        decoded = decode_token()
+        if decoded == None:
```
samuzzal-choudhury: @fridex good catch. Actually, this check is present in login_required. Need to check why that did not work.

bayesian/utils.py
samuzzal-choudhury: `server_create_component_bookkeeping` function is used only for bookkeeping component-analyses.

bayesian/auth.py
fridex: @samuzzal-choudhury OK, now I see. Thanks for clarification!

bayesian/utils.py
samuzzal-choudhury: @fridex I am not repeating the call to server_create_component_bookkeeping. There are two scenarios for which server_create_component_bookkeeping is called separately.
1.  When component data is available on graphdb.
2. When component data is not available, componentApiFlow is called before calling bayesianFlow or bayesianApiFlow.

Did I miss anything here?

fridex: I meant the following line:

```python
MavenCoordinates.normalize_str(package) if ecosystem == 'maven' else package
```

In other words you can directly pass:

```python
args = {
  'name': component,
   ...
}

```

samuzzal-choudhury:
Closing this PR in favour of the new PR https://github.com/fabric8-analytics/fabric8-analytics-server/pull/80

Issue #64, opened 2017-07-11
Component search response contains a string with JSON inside


tisnik:
If I access the following endpoint:
```
http://localhost:32000/api/v1/component-search/foobar
```

I get:
```
"{\"result\": []}"
```

I'd assume the following response:
```
{"result": []}
```

I.e. it seems that the search result is explicitly converted to JSON two times

PR #65, opened 2017-07-11
Do not jsonify dict with search results

fridex:
Fixes: #64


tisnik:
Looks ok, thank you very much @fridex. I'd also suggest to do the same change 13 lines above (for no results).

fridex:
> I'd also suggest to do the same change 13 lines above (for no results).

Nice spot! Totally missed that one. Should be fixed now :)

tisnik:
Nice! TYVM. I'll close it after CI completion ;)

PR #69, opened 2017-07-13
Sync update_selinon.sh script with worker

msrb:


msrb:
Testing locally first...

PR #84, opened 2017-07-21
Better disable authentication testing

tisnik:
in the docker_compose.yml, the setting of DISABLE_AUTHENTICATION to 0 nor "0" nor "" does not work because the code actually test the existence of this variable, not the content

### DIFF
--- a/bayesian/auth.py
+++ b/bayesian/auth.py
@@ -38,7 +38,7 @@ def login_required(view):
     # being able to tail logs and see if stuff is going fine
     def wrapper(*args, **kwargs):
         # Disable authentication for local setup
-        if getenv('DISABLE_AUTHENTICATION'):
+        if getenv('DISABLE_AUTHENTICATION') in ('1', 'True', 'true'):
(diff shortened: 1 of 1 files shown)


tisnik:
[test this]

fridex:
BTW worth to unify this across services - jobs service has different idion https://github.com/fabric8-analytics/fabric8-analytics-jobs/blob/master/f8a_jobs/defaults.py#L18

tisnik:
@fridex I'll look at f8a_jobs as well as I feel it is better to support all usual options - numeric 1/0 and booleans True/False, true/false as well (our yml files are a bit inconsistent here)

fridex:
> @fridex I'll look at f8a_jobs as well as I feel it is better to support all usual options - numeric 1/0 and booleans True/False, true/false as well (our yml files are a bit inconsistent here)

Not needed anymore, I've already done that in https://github.com/fabric8-analytics/fabric8-analytics-jobs/pull/106

PR #89, opened 2017-07-21
Fix NPE when authentication is disabled

tisnik:
When authentication is disabled (devels machines etc.) then the user_profile can't be decoded from the token -> the service ends with NPE on user_profile.get('email', ...)

This might fix this issue


tisnik:
The default devel configuration has authentication disabled. The same was true for the integration tests, thought it has been fixed via https://github.com/fabric8-analytics/fabric8-analytics-common/pull/133

I'm not against forcing to use authentication in all cases, but the code needs to be cleaned up as well (the non-auth part won't be covered by tests etc. etc.)

bayesian/utils.py
```diff
@@ -32,12 +32,19 @@ def server_run_flow(flow_name, flow_args):
...
+    else:
+        return default_email
+
 def server_create_component_bookkeeping(ecosystem, name, version, user_profile):
     args = {
         'external_request_id': uuid.uuid4().hex,
         'data': {
             'api_name': 'component_analyses',
-            'user_email': user_profile.get('email','bayesian@redhat.com'),
+            'user_email': get_user_email(user_profile),
```
fridex: A shorter version could be:

```python
(user_profile or {}).get('email', 'bayesian@redhat.com')
```

fridex:
> Can you please describe the scenario how and why we would like to turn authentication off?

I think the main use case is local setup where we don't want to bother with authentication at all when testing some code changes during development (see [local setup configuration](https://github.com/fabric8-analytics/fabric8-analytics-common/blob/master/docker-compose.yml#L73)) Also if we decide to deploy on some internal cluster (internal use-case), we should be fine to turn off authentication.

Anyway - this does not break anything but rather fixes issues we are experiencing, @samuzzal-choudhury what do you think about getting this in?

tuxdna:
Code looks good to me. Let me test it out.

samuzzal-choudhury:
@tisnik could you please resolve the conflicts?

tuxdna:
@tisnik I have resolved conflicts in my fork https://github.com/tuxdna/fabric8-analytics-server/tree/fix-NPE-when-authentication-is-disabled , and the code works fine for me. You may want to pull from there and include it in this PR.

tuxdna:
@samuzzal-choudhury  Once the tests pass, we can merge this.

PR #96, opened 2017-07-27
add mechanism to trace pkg with no sentiment details

SaketJNU:
Hi,
This PR is created to resolve the condition, when we have no sentiment details for user_stack analysis, which is generating nontype attribute, details of the error is given in : http://pastebin.test.redhat.com/504244 

However, this is not the complete solution, but it avoids the error this time, and requires further enhancement.
Thanks,
Saket

### DIFF
--- a/bayesian/api_v1.py
+++ b/bayesian/api_v1.py
@@ -342,45 +342,44 @@ class StackAnalysesGETV2(ResourceWithSchema):
         manifest_response = []
         recommendation = {}
 
-        if stack_result != None and 'task_result' in stack_result:
+        if stack_result != None:
             if stack_result["task_result"] != None:
                 started_at = stack_result["task_result"]["_audit"]["started_at"]
                 finished_at = stack_result["task_result"]["_audit"]["ended_at"]
 
         if reco_result is not None and 'task_result' in reco_result:
-                if reco_result["task_result"] != None:
-                    recommendation = reco_result['task_result']['recommendations']
+            if reco_result["task_result"] != None:
+                recommendation = reco_result['task_result']['recommendations']
 
         # Populate sentiment score for packages in user's stack
         if stack_result is not None:
-            user_stack_deps = stack_result.get('task_result').get('user_stack_info').get('dependencies',[])
-            if user_stack_deps:
-                for dep in user_stack_deps:
-                    if user_stack_sentiment_result is not None:
-                        if user_stack_sentiment_result.get('task_result').get(dep['name']) is not None:
-                            dep['sentiment']['overall_score'] = user_stack_sentiment_result.get('task_result').\
-                                get(dep['name']).get('score', 0)
-                        else:
-                            dep['sentiment'] = {}
+            user_stack_deps = stack_result['task_result']['user_stack_info'].get('dependencies',[])
+            for dep in user_stack_deps:
+                if user_stack_sentiment_result is not None:
+                    user_stack_sentiment = user_stack_sentiment_result.get('task_result')
+                    print("i am user_stack", user_stack_sentiment)
+                    if user_stack_sentiment.get(dep['name']) is not None:
+                        dep['sentiment']['overall_score'] = user_stack_sentiment.get(dep['name']).get('score', 0)
+                    else:
+                        dep['sentiment'] = {}
 
         # Populate sentiment score for recommended packages
         if reco_result is not None:
             alternate = reco_result.get('task_result').get('recommendations').get('alternate', [])
-            if alternate:
-                for pkg in alternate:
-                    if reco_pkg_sentiment_result.get("task_result") is not None:
-                        if reco_pkg_sentiment_result.get('task_result').get(dep['name']) is not None:
-                            pkg['sentiment']['overall_score'] = reco_pkg_sentiment_result.get('task_result').\
-                                get(dep['name']).get('score', 0)
-                        else:
-                            pkg['sentiment'] = {}
+            for pkg in alternate:
+                if reco_pkg_sentiment_result is not None:
+                    reco_pkg_sentiment_alter = reco_pkg_sentiment_result.get('task_result')
+                    if reco_pkg_sentiment_alter is not None:
+                        pkg['sentiment']['overall_score'] = reco_pkg_sentiment_alter.get(pkg['name'], {}).get('score', 0)
+                    else:
+                        pkg['sentiment'] = {}
 
             companion = reco_result.get('task_result').get('recommendations').get('companion',[])
-
-            if companion:
-                for pkg in companion:
-                    if reco_pkg_sentiment_result['task_result'] is not None:
-                        pkg['sentiment']['overall_score'] = reco_pkg_sentiment_result['task_result'].get(pkg['name']).get('score', 0)
+            for pkg in companion:
+                if reco_pkg_sentiment_result is not None:
+                    reco_pkg_sentiment_companion = reco_pkg_sentiment_result.get('task_result')
+                    if reco_pkg_sentiment_companion is not None:
+                        pkg['sentiment']['overall_score'] = reco_pkg_sentiment_companion.get(pkg['name'], {}).get('score', 0)
                     else:
                         pkg['sentiment'] = {}
 
--- a/bayesian/api_v1.py
+++ b/bayesian/api_v1.py
@@ -349,39 +349,52 @@ class StackAnalysesGETV2(ResourceWithSchema):
 
         if reco_result is not None and 'task_result' in reco_result:
             if reco_result["task_result"] != None:
-                recommendation = reco_result['task_result']['recommendations']
+                recommendation = reco_result.get('task_result').get('recommendations', {})
 
         # Populate sentiment score for packages in user's stack
         if stack_result is not None:
-            user_stack_deps = stack_result['task_result']['user_stack_info'].get('dependencies',[])
+            user_stack_deps = stack_result.get('task_result', {}).get('user_stack_info', {}).get('dependencies',[])
             for dep in user_stack_deps:
                 if user_stack_sentiment_result is not None:
-                    user_stack_sentiment = user_stack_sentiment_result.get('task_result')
-                    print("i am user_stack", user_stack_sentiment)
+                    user_stack_sentiment = user_stack_sentiment_result.get('task_result', {})
                     if user_stack_sentiment.get(dep['name']) is not None:
                         dep['sentiment']['overall_score'] = user_stack_sentiment.get(dep['name']).get('score', 0)
+                        dep['sentiment']['magnitude'] = user_stack_sentiment.get(dep['name'], {}).get('magnitude', 0)
                     else:
-                        dep['sentiment'] = {}
+                        dep['sentiment'] = {
+                            "latest_comment": "",
+                            "overall_score": 0,
+                            "magnitude": 0
+                        }
 
         # Populate sentiment score for recommended packages
         if reco_result is not None:
-            alternate = reco_result.get('task_result').get('recommendations').get('alternate', [])
+            alternate = reco_result.get('task_result', {}).get('recommendations', {}).get('alternate', [])
             for pkg in alternate:
                 if reco_pkg_sentiment_result is not None:
-                    reco_pkg_sentiment_alter = reco_pkg_sentiment_result.get('task_result')
+                    reco_pkg_sentiment_alter = reco_pkg_sentiment_result.get('task_result', {})
                     if reco_pkg_sentiment_alter is not None:
                         pkg['sentiment']['overall_score'] = reco_pkg_sentiment_alter.get(pkg['name'], {}).get('score', 0)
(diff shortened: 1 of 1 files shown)


bayesian/api_v1.py
```diff
@@ -347,35 +347,42 @@ def get(external_request_id):
                 started_at = stack_result["task_result"]["_audit"]["started_at"]
                 finished_at = stack_result["task_result"]["_audit"]["ended_at"]
 
-
         if reco_result is not None and 'task_result' in reco_result:
-            if reco_result["task_result"] != None:
-                recommendation = reco_result['task_result']['recommendations']
+                if reco_result["task_result"] != None:
```
harjinder-hari: extra indentation.

bayesian/api_v1.py
```diff
@@ -347,35 +347,42 @@ def get(external_request_id):
...
+                    recommendation = reco_result['task_result']['recommendations']
```
harjinder-hari: extra indentation

bayesian/api_v1.py
```diff
@@ -347,39 +347,43 @@ def get(external_request_id):
...
         # Populate sentiment score for recommended packages
         if reco_result is not None:
             alternate = reco_result.get('task_result').get('recommendations').get('alternate', [])
             if alternate:
                 for pkg in alternate:
                     if reco_pkg_sentiment_result.get("task_result") is not None:
-                        pkg['sentiment']['overall_score'] = reco_pkg_sentiment_result.get('task_result').get(dep['name']).get('score', 0)
-                    else:
-                        pkg['sentiment'] = {}
+                        if reco_pkg_sentiment_result.get('task_result').get(dep['name']) is not None:
```
miteshvp: if reco_pkg_sentiment_result.get('task_result', {}).get(dep['name'], {}):

bayesian/api_v1.py
```diff
@@ -347,39 +347,43 @@ def get(external_request_id):
...
         # Populate sentiment score for packages in user's stack
         if stack_result is not None:
             user_stack_deps = stack_result.get('task_result').get('user_stack_info').get('dependencies',[])
             if user_stack_deps:
                 for dep in user_stack_deps:
                     if user_stack_sentiment_result is not None:
-                        dep['sentiment']['overall_score'] = user_stack_sentiment_result.get('task_result').get(dep['name']).get('score', 0)
-                    else:
-                        dep['sentiment'] = {}
+                        if user_stack_sentiment_result.get('task_result').get(dep['name']) is not None:
```
miteshvp: if user_stack_sentiment_result.get('task_result', {}).get(dep['name'], {}):

bayesian/api_v1.py
```diff
@@ -347,35 +347,42 @@ def get(external_request_id):
...
 
         # Populate sentiment score for packages in user's stack
         if stack_result is not None:
-            user_stack_deps = stack_result['task_result']['user_stack_info']['dependencies']
-            for dep in user_stack_deps:
-                if user_stack_sentiment_result is not None:
-                    dep['sentiment']['overall_score'] = user_stack_sentiment_result['task_result'][dep['name']]['score']
-                else:
-                    dep['sentiment'] = {}
+            user_stack_deps = stack_result.get('task_result').get('user_stack_info').get('dependencies',[])
```
harjinder-hari: if default value is not specified, then we can use subscription method to fetch a value for given key. For example, here we can say `stack_result['task_result']['user_stack_info'].get('dependencies', [])` as this is more readable.

bayesian/api_v1.py
```diff
@@ -347,39 +347,43 @@ def get(external_request_id):
...
+
             if companion:
                 for pkg in companion:
                     if reco_pkg_sentiment_result['task_result'] is not None:
                         pkg['sentiment']['overall_score'] = reco_pkg_sentiment_result['task_result'].get(pkg['name']).get('score', 0)
                     else:
                         pkg['sentiment'] = {}
 
+            if reco_result is not None and 'task_result' in reco_result:
+                if reco_result["task_result"] != None:
```
miteshvp: if reco_result["task_result"] is not None:

bayesian/api_v1.py
```diff
@@ -347,35 +347,42 @@ def get(external_request_id):
...
         # Populate sentiment score for packages in user's stack
         if stack_result is not None:
-            user_stack_deps = stack_result['task_result']['user_stack_info']['dependencies']
-            for dep in user_stack_deps:
-                if user_stack_sentiment_result is not None:
-                    dep['sentiment']['overall_score'] = user_stack_sentiment_result['task_result'][dep['name']]['score']
-                else:
-                    dep['sentiment'] = {}
+            user_stack_deps = stack_result.get('task_result').get('user_stack_info').get('dependencies',[])
+            if user_stack_deps:
```
harjinder-hari: This `if` is not required as you have already made sure that `user_stack_deps` will never be None. In the worst case, it will be an empty list.

bayesian/api_v1.py
```diff
@@ -347,35 +347,42 @@ def get(external_request_id):
...
+                for dep in user_stack_deps:
+                    if user_stack_sentiment_result is not None:
+                        if user_stack_sentiment_result.get('task_result').get(dep['name']) is not None:
```
harjinder-hari: The value of `user_stack_sentiment_result.get('task_result')` can be kept in a new variable e.g. `user_stack_sentiment` so that we can avoid fetching the `task_result` value again and again. This is also more readable.

bayesian/api_v1.py
```diff
@@ -347,35 +347,42 @@ def get(external_request_id):
...
         # Populate sentiment score for recommended packages
         if reco_result is not None:
-            alternate = reco_result['task_result']['recommendations']['alternate']
-            for pkg in alternate:
-                if reco_pkg_sentiment_result is not None:
-                    pkg['sentiment']['overall_score'] = reco_pkg_sentiment_result['task_result'][pkg['name']]['score']
-                else:
-                    pkg['sentiment'] = {}
+            alternate = reco_result.get('task_result').get('recommendations').get('alternate', [])
+            if alternate:
```
harjinder-hari: This `if` is not required as you have already made sure that `alternate ` will never be None. In the worst case, it will be an empty list.

bayesian/api_v1.py
```diff
@@ -347,35 +347,42 @@ def get(external_request_id):
...
+                for pkg in alternate:
+                    if reco_pkg_sentiment_result.get("task_result") is not None:
```
harjinder-hari: `reco_pkg_sentiment_result` might be None ... we need to check for that just like you checked for `user_stack_sentiment_result`

Also, value of `reco_pkg_sentiment_result.get('task_result')` can be kept in a new variable e.g. `reco_pkg_sentiment` so that we can avoid fetching the `task_result` value again and again. This is also more readable.

bayesian/api_v1.py
```diff
@@ -347,35 +347,42 @@ def get(external_request_id):
...
 
-            companion = reco_result['task_result']['recommendations']['companion']
-            for pkg in companion:
-                if reco_pkg_sentiment_result is not None:
-                    pkg['sentiment']['overall_score'] = reco_pkg_sentiment_result['task_result'][pkg['name']]['score']
-                else:
-                    pkg['sentiment'] = {}
+            companion = reco_result.get('task_result').get('recommendations').get('companion',[])
+
+            if companion:
```
harjinder-hari: This `if` is not required as you have already made sure that `companion ` will never be None. In the worst case, it will be an empty list.

bayesian/api_v1.py
```diff
@@ -347,35 +347,42 @@ def get(external_request_id):
...
+                for pkg in companion:
+                    if reco_pkg_sentiment_result['task_result'] is not None:
```
harjinder-hari: `reco_pkg_sentiment_result` might be None ... we need to check for that just like you checked for `user_stack_sentiment_result`

Also, value of `reco_pkg_sentiment_result.get('task_result')` can be kept in a new variable e.g. `reco_pkg_sentiment` so that we can avoid fetching the task_result value again and again. This is also more readable.

bayesian/api_v1.py
SaketJNU: fixed the extra indentation

bayesian/api_v1.py
SaketJNU: Here motivation was to keep same track for all keys. However fixed the changes are per suggestion.

bayesian/api_v1.py
SaketJNU: agreed with it and removed the mentioned 'if' statement.

bayesian/api_v1.py
SaketJNU: Updated by including the new variable

bayesian/api_v1.py
SaketJNU: fixed this comment as per suggestion

bayesian/api_v1.py
SaketJNU: Updated by including the new variable.

bayesian/api_v1.py
SaketJNU: removed the if statement

bayesian/api_v1.py
SaketJNU: updated the if condition and added the new check point as suggestion.

harjinder-hari: @SaketJNU, Thanks for addressing the comments. Can you please push the commit. Your latest changes are not visible yet.

harjinder-hari:
@SaketJNU, The changes look good to me. Thanks for addressing my comments.

bayesian/api_v1.py
```diff
@@ -342,40 +342,46 @@ def get(external_request_id):
...
-                else:
-                    pkg['sentiment'] = {}
+                    reco_pkg_sentiment_alter = reco_pkg_sentiment_result.get('task_result')
+                    if reco_pkg_sentiment_alter is not None:
+                        pkg['sentiment']['overall_score'] = reco_pkg_sentiment_alter.get(pkg['name'], {}).get('score', 0)
+                    else:
+                        pkg['sentiment'] = {}
 
-            companion = reco_result['task_result']['recommendations']['companion']
+            companion = reco_result.get('task_result').get('recommendations').get('companion',[])
```
miteshvp: This wil still fail if there is no task_result block or recommendations block.

bayesian/api_v1.py
```diff
@@ -342,40 +342,46 @@ def get(external_request_id):
...
+                    print("i am user_stack", user_stack_sentiment)
+                    if user_stack_sentiment.get(dep['name']) is not None:
+                        dep['sentiment']['overall_score'] = user_stack_sentiment.get(dep['name']).get('score', 0)
+                    else:
+                        dep['sentiment'] = {}
 
         # Populate sentiment score for recommended packages
         if reco_result is not None:
-            alternate = reco_result['task_result']['recommendations']['alternate']
+            alternate = reco_result.get('task_result').get('recommendations').get('alternate', [])
```
miteshvp: will fail if no task_result or recommendations block is available

bayesian/api_v1.py
```diff
@@ -342,40 +342,46 @@ def get(external_request_id):
...
 
-
         if reco_result is not None and 'task_result' in reco_result:
             if reco_result["task_result"] != None:
                 recommendation = reco_result['task_result']['recommendations']
 
         # Populate sentiment score for packages in user's stack
         if stack_result is not None:
-            user_stack_deps = stack_result['task_result']['user_stack_info']['dependencies']
+            user_stack_deps = stack_result['task_result']['user_stack_info'].get('dependencies',[])
```
miteshvp: same as below comments

bayesian/api_v1.py
```diff
@@ -342,40 +342,46 @@ def get(external_request_id):
...
             for dep in user_stack_deps:
                 if user_stack_sentiment_result is not None:
-                    dep['sentiment']['overall_score'] = user_stack_sentiment_result['task_result'][dep['name']]['score']
-                else:
-                    dep['sentiment'] = {}
+                    user_stack_sentiment = user_stack_sentiment_result.get('task_result')
+                    print("i am user_stack", user_stack_sentiment)
```
miteshvp: print - unwanted

SaketJNU: removed

bayesian/api_v1.py
SaketJNU: updated

bayesian/api_v1.py
SaketJNU: updated

bayesian/api_v1.py
SaketJNU: fixed this

bayesian/api_v1.py
SaketJNU: updated

SaketJNU:
In the latest commit a new attribute  "magnitude" in sentiment is also added which will be useful to show sentiment score in UI, i.e. if magnitude is 0 that means we do't have data in stackoverflow to compute sentiment score. So in this case not to sentiment score to end user. 
cc: @samuzzal-choudhury

PR #122, opened 2017-08-10
Stop using selinon hacks

pkajaba:
related to: https://github.com/fabric8-analytics/fabric8-analytics-worker/pull/233


pkajaba:
I have to update version of worker too, so this can be merge after changes in worker are done

PR #128, opened 2017-08-24
remove sentiment-scorer from workflow.

SaketJNU:
An action item for https://github.com/openshiftio/openshift.io/issues/645

### DIFF
--- a/bayesian/api_v1.py
+++ b/bayesian/api_v1.py
@@ -371,6 +371,11 @@ class StackAnalysesGET(ResourceWithSchema):
                 else:
                     dep["topic_list"] = []
 
+        for stack in stacks:
+            stack["recommendation"] = get_item_from_list_by_key_value(
+                recommendations,
+                "manifest_file_path",
+                stack.get("manifest_file_path"))
             manifest_response.append(stack)
 
         return {
--- a/bayesian/api_v1.py
+++ b/bayesian/api_v1.py
@@ -317,10 +317,6 @@ class StackAnalysesGET(ResourceWithSchema):
 
         stack_result = retrieve_worker_result(rdb, external_request_id, "stack_aggregator_v2")
         reco_result = retrieve_worker_result(rdb, external_request_id, "recommendation_v2")
-        user_stack_sentiment_result = retrieve_worker_result(rdb, external_request_id,
-                                                             "user_stack_sentiment_scorer")
-        reco_pkg_sentiment_result = retrieve_worker_result(rdb, external_request_id,
-                                                           "reco_pkg_sentiment_scorer")
 
         if stack_result is None and reco_result is None:
             raise HTTPError(202, "Analysis for request ID '{t}' is in progress".format(
@@ -352,7 +348,6 @@ class StackAnalysesGET(ResourceWithSchema):
             recommendations = reco_result.get("task_result", {}).get("recommendations",
                                                                      recommendations)
 
-        # Populate sentiment score for packages in user's stack
         if not stacks:
             return {
                 "version": version,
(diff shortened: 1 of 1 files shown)


SaketJNU:
An attempted is made is to remove "sentiment-scorer" from entire places. Thanks

bayesian/api_v1.py
```diff
@@ -403,82 +398,7 @@ def get(external_request_id):
...
-                # Adding topics from the recommendations
-                stack_recommendation = get_item_from_list_by_key_value(recommendations,
-                                                                       "manifest_file_path",
-                                                                       stack.get(
-                                                                           "manifest_file_path"))
-                if stack_recommendation is not None:
-                    dep["topic_list"] = stack_recommendation.get("input_stack_topics",
-                                                                 {}).get(dep.get('name'), [])
-                else:
-                    dep["topic_list"] = []
```
jpopelka: Do we really want to remove also this populating of `topic_list` part ?

If not, then I'd suggest to move the `stack_recommendation = get_item_from_list_by_key_value(...)` outside the `for dep in user_stack_deps:` loop as it's invariant to it.

bayesian/api_v1.py
```diff
@@ -403,82 +398,7 @@ def get(external_request_id):
...
-                            "overall_score": 0,
-                            "magnitude": 0
-                        }
-
-        for stack in stacks:
-            stack["recommendation"] = get_item_from_list_by_key_value(
-                recommendations,
-                "manifest_file_path",
-                stack.get("manifest_file_path"))
+            stack["recommendation"] = get_item_from_list_by_key_value(recommendations, "manifest_file_path", stack.get("manifest_file_path"))
```
jpopelka: The line is too long. Keep it as it is now please.

SaketJNU: reverted to its previous style.

bayesian/api_v1.py
SaketJNU: we need ```dep[topic_list]``` so keeping this. Now, I also tried to update the suggested line and removed from the above-mentioned for loop. One more thing, i saw repetition of three lines of codes with  ```stack_recommendation = get_item_from_list_by_key_value(...)```, removed the repeated lines and appended one additional line with the updated codes. hopefully, it will have the same effect. Thanks

jpopelka: I think we shouldn't simply remove the setting of `stack["recommendation"]` (as you do at the end of the `get()` method).

So what do you think about replacing  every occurrence of `stack_recommendation` with `stack["recommendation"]` ?

jpopelka:
It'd be nice if you could rebase your changes to get rid of the `Merge branch 'master' of ...` commit

jpopelka:
Have you seen [this comment](https://github.com/fabric8-analytics/fabric8-analytics-server/pull/128#discussion_r142370640) ?
I think we should replace `stack_recommendation` with `stack["recommendation"]`.

jpopelka:
> It'd be nice if you could rebase your changes

With `rebase` I mean just to run `git rebase master` - that should be enough to remove the `Merge branch 'master' of ...` commits.

You don't need to `squash` the commits into one (as you did).
You can, but in that case please give the commit some meaningful comment - `'reabse the branch'` doesn't say anything.

bayesian/api_v1.py
SaketJNU: I did the earlier change for ```stack_recommendation``` and ```stack["recommendation"]``` which is wrong. Both are different entities, and we should not touch them. So reverting these changes.

jpopelka:
[test]

jpopelka:
I'm sorry for the constant nitpicking, but you named the commit where we remove all the sentiment-scorer code: `'append topic_list'`, which is not what it does - it does not append topic_list, it removes all the sentiment-scorer code.

jpopelka:
[test]

Issue #139, opened 2017-08-31
RFE: the API endpoint to force analysis for selected component on given ecosystem


tisnik:
Currently, the following endpoint called with GET method starts analysis for selected component+version only when the analysis is not finished already:

```
/api/v1/component-analyses/<ecosystem>/<package>/<version>
```

It would be nice to have the same endpoint, but for the method POST or PUT to force new analysis, even if old analysis already exists.

fridex:
An idea for next iteration - we could make it more "sensitive" by providing endpoint for updating only certain task results (as we already have selective task runs). A user could say - I want to update only github data and so. This would be similar to https://github.com/fabric8-analytics/fabric8-analytics-server/issues/38

tisnik:
It would be really nice @fridex, TY in advance :)

samuzzal-choudhury:
This issue was moved to openshiftio/openshift.io#853

PR #141, opened 2017-08-31
Force component analyses run on POST

fridex:
Fixes: #139

### DIFF
--- a/bayesian/api_v1.py
+++ b/bayesian/api_v1.py
@@ -274,6 +274,17 @@ class ComponentAnalyses(ResourceWithSchema):
                                                                                      package=package, version=version)
             raise HTTPError(404, msg)
 
+    @staticmethod
+    def post(ecosystem, package, version):
+        decoded = decode_token()
+        if ecosystem == 'maven':
+            package = MavenCoordinates.normalize_str(package)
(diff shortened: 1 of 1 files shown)


fridex:
CC @tisnik

bayesian/api_v1.py
```diff
@@ -274,6 +274,17 @@ def get(ecosystem, package, version):
...
 
+    @staticmethod
+    def post(ecosystem, package, version):
+        decoded = decode_token()
+        if ecosystem == 'maven':
+            package = MavenCoordinates.normalize_str(package)
+        package = case_sensitivity_transform(ecosystem, package)
+
+        server_create_analysis(ecosystem, package, version,
+                               user_profile=decoded or {}, api_flow=True, force=True, force_graph_sync=False)
```
fridex: @samuzzal-choudhury it looks like I'm getting error on local setup as there is no token, `decoded` is `None` and running the flow fails as there are not tracked user_profile information (thus `decoded or None` part). Is this intentional? Otherwise we would need to fix it on other places. WDYT?

tisnik:
It looks ok,  @fridex 🥇

humaton:
LGTM merging

PR #143, opened 2017-08-31
fixes topics not shown for user stack components

miteshvp:


fridex:
Will wait for CI to give it green not to break anything. We can merge it after that.

PR #148, opened 2017-09-15
Run Python linter on Jenkins

tisnik:
Run Python linter on Jenkins (CI), but don't fail ATM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ __pycache__/
 # Distribution / packaging
 .Python
 env/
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -27,6 +27,12 @@ node('docker') {
         }
     }
 
+    stage('Source code linter') {
```
pkajaba: This won't work since we don't you Jenkinsfile anymore. You have to update Makefile,

PR #163, opened 2017-09-21
Improve couple functions

jpopelka:
retrieve_worker_result():

- the 'for' loop has no sense since it overwrites 'result' in each step
- use query.one() since we are interested only in one result
- return None if there's 0 or >1 matching results
- rollback on general SQLAlchemyError

get_request_count():

- rollback on SQLAlchemyError


pkajaba:
LGTM! Since main CI job succeed we can go with a merge.

PR #165, opened 2017-09-26
Changed the logic of stack analysis using a github url

samuzzal-choudhury:
The existing implementation of performing stack analysis by means of a GitHub URL as parameter clones the repo, finds the manifest files and then extracts the content of those files to be stored in the RDS against the request ID.

The use of GitHub clone results in regular exceptions mentioned in the issue https://github.com/openshiftio/openshift.io/issues/814.

This PR addresses the problem in another way:
1.  Fetch the SHA for the last commit of the repo.
2.  Fetch the content tree recursively to fetch all the files with their corresponding directories.
3.  Find out the manifest file location from the fetched tree and generate the raw contents URL for those files.
4.  Download those manifest file contents from the generated GitHub raw contents URL.

This solution is aimed to work on any platform by standard GitHub API endpoint calls.

### DIFF
--- a/bayesian/api_v1.py
+++ b/bayesian/api_v1.py
@@ -652,7 +652,7 @@ class StackAnalysesV2(ResourceWithSchema):
         decoded = decode_token()
         github_url = request.form.get("github_url")
         if github_url is not None:
-            files = GithubRead().get_manifest_details(github_url)
+            files = GithubRead().get_files_github_url(github_url)
         else:
             files = request.files.getlist('manifest[]')
             filepaths = request.values.getlist('filePath[]')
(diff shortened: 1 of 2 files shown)


msrb:
> This solution is aimed to work on any platform by standard GitHub API endpoint calls.

Are there any limits that we might hit while interacting with the API? I think unauthenticated requests are limited to 60 per hour (not 100% sure about this).

msrb:
[test]

samuzzal-choudhury:
Looking at the alternatives after a call with @msrb

jyasveer:
This is a cleaner solution, where we do not need to clone the repo.

humaton:
@samuzzal-choudhury  Are there any updates on this?

samuzzal-choudhury:
@humaton  Post discussions with @msrb we thought that might be a permissions issue on prod. Moreover, GitHub API come with rate limits which is a potential problem in my proposed solution. Do you have anything in mind for this?

humaton:
Not really, I was just going through opened PRs to see the current state.

msrb:
@samuzzal-choudhury looks like this PR needs to be rebased.

samuzzal-choudhury:
Obsolete change. Hence, closing

PR #167, opened 2017-09-27
Create: User Intent API

jyasveer:
Create an api for the crowd sourcing user story https://github.com/openshiftio/openshift.io/issues/1002

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,4 +80,4 @@ The marked offline tests can then be run as:
 If the Docker container based tests have been run, then this may complain
 about mismatched locations in compiled files. Those can be deleted using:
 
-    (bayesian) $ sudo find -name *.pyc -delete 
+    (bayesian) $ sudo find -name *.pyc -delete
--- a/README.md
+++ b/README.md
@@ -80,4 +80,4 @@ The marked offline tests can then be run as:
 If the Docker container based tests have been run, then this may complain
 about mismatched locations in compiled files. Those can be deleted using:
 
-    (bayesian) $ sudo find -name *.pyc -delete
+    (bayesian) $ sudo find -name *.pyc -delete 
--- a/bayesian/api_v1.py
+++ b/bayesian/api_v1.py
@@ -415,6 +415,37 @@ class UserFeedback(ResourceWithSchema):
 
         return {'status': 'success'}
 
+class UserIntent(ResourceWithSchema):
+    method_decorators = [login_required]
+    _ANALYTICS_BUCKET_NAME = "{}-{}".format(
+        os.environ.get('DEPLOYMENT_PREFIX', 'unknown'),
+        os.environ.get("AWS_ANALYTICS_BUCKET", "bayesian-user-intent"))
+
+    @staticmethod
+    def post():
+        input_json = request.get_json()
+
+        if not input_json:
+            raise HTTPError(400, error="Expected JSON request")
+        
+        if 'component' not in input_json:
+            raise HTTPError(400, error="Expected component name in the request")
+        
+        if 'intent' not in input_json:
+            raise HTTPError(400, error="Expected intent in the request")
+
+        s3 = AmazonS3(bucket_name=UserIntent._ANALYTICS_BUCKET_NAME)
+        s3.connect()
+
+        # Store data
+        key = "{}".format(input_json["component"])
+        if s3.object_exists(key):
+            existingObject = s3.retrieve_dict(key)
+            existingIntentList = existingObject.get('intent', [])
+            input_json.get('intent', []).extend(existingIntentList)
+        s3.store_dict(input_json, key)
+
+        return {'status': 'success'}
 
 class StackAnalyses(ResourceWithSchema):
     method_decorators = [login_required]
@@ -596,6 +627,7 @@ add_resource_no_matter_slashes(SystemVersion, '/system/version')
 add_resource_no_matter_slashes(StackAnalyses, '/stack-analyses')
 add_resource_no_matter_slashes(StackAnalysesGET, '/stack-analyses/<external_request_id>')
 add_resource_no_matter_slashes(UserFeedback, '/user-feedback')
+add_resource_no_matter_slashes(UserIntent, '/user-intent')
 add_resource_no_matter_slashes(PublishedSchemas, '/schemas')
 add_resource_no_matter_slashes(PublishedSchemas, '/schemas/<collection>',
                                endpoint='get_schemas_by_collection')
--- a/bayesian/api_v1.py
+++ b/bayesian/api_v1.py
@@ -422,10 +422,10 @@ class UserIntent(ResourceWithSchema):
 
     @staticmethod
     def get(user, ecosystem):
-        if user is None:
+        if not user:
             raise HTTPError(400, error="Expected user name in the request")
 
-        if ecosystem is None:
+        if not ecosystem:
             raise HTTPError(400, error="Expected ecosystem in the request")
 
         s3 = StoragePool.get_connected_storage('S3ManualTagging')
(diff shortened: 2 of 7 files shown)


bayesian/api_v1.py
```diff
@@ -518,6 +518,37 @@ def post():
...
+            raise HTTPError(400, error="Expected intent in the request")
+
+        s3 = AmazonS3(bucket_name=UserIntent._ANALYTICS_BUCKET_NAME)
+        s3.connect()
+
+        # Store data
+        key = "{}".format(input_json["component"])
+        if s3.object_exists(key):
+            existingObject = s3.retrieve_dict(key)
+            existingIntentList = existingObject.get('intent', [])
```
fridex: Please do not mix camel case and snake case. We use snake case for variables.

jyasveer: no problem. i will modify this.

jyasveer:
@fridex What do you mean by adapters here ? And how does it help here ? can you please explain ? Thanks

pkajaba:
[test]

fridex:
@jyasveer added selinon initialization that should fix issues when getting storage adapter in 
faa3bb3

fridex:
[test]

fridex:
[test]

fridex:
@jyasveer please check CI output, pylint is complaining:

```
bayesian/__init__.py
    Pass
bayesian/api_v1.py
bayesian/api_v1.py:53:1: E302 expected 2 blank lines, found 1
bayesian/api_v1.py:418:1: E302 expected 2 blank lines, found 1
bayesian/api_v1.py:424:101: E501 line too long (103 > 100 characters)
bayesian/api_v1.py:425:1: W293 blank line contains whitespace
bayesian/api_v1.py:433:1: W293 blank line contains whitespace
bayesian/api_v1.py:440:1: W293 blank line contains whitespace
bayesian/api_v1.py:444:1: W293 blank line contains whitespace
bayesian/api_v1.py:452:1: W293 blank line contains whitespace
bayesian/api_v1.py:463:1: E302 expected 2 blank lines, found 1
    Fail
```

jyasveer:
@fridex please check

bayesian/api_v1.py
```diff
@@ -416,6 +417,47 @@ def post():
         return {'status': 'success'}
 
 
+class UserIntent(ResourceWithSchema):
+    method_decorators = [login_required]
+
+    @staticmethod
+    def get(user, ecosystem):
+        if user is not None and ecosystem is not None:
```
fridex: This looks like a logical error. What do you want to accomplish with this check?

bayesian/api_v1.py
```diff
@@ -416,6 +417,47 @@ def post():
...
+            raise HTTPError(404, "Invalid user '{u}' and ecosystem '{e}'.".format(u=user,
+                                                                                  e=ecosystem))
+
+        s3 = StoragePool.get_connected_storage('S3ManualTagging')
+        # get user data
+        return s3.fetch_user_data(user, ecosystem)
```
fridex: The adapter raises `botocore.exceptions.ClientError` if the given user/ecosystem is not present. It will result in internal server error. It would be nice to capture such exception and return 404 instead of 500.

bayesian/api_v1.py
```diff
@@ -416,6 +417,47 @@ def post():
...
+
+    @staticmethod
+    def post():
+        input_json = request.get_json()
+
+        if not input_json:
+            raise HTTPError(400, error="Expected JSON request")
+
+        if 'manual_tagging' not in input_json:
```
fridex: This looks like using this endpoint for two different purposes. Do we really want to change endpoint behaviour based on JSON that is sent? This might be confusing.

jyasveer: yes @fridex this is the requirement. where the same end point will be used for manual tagging and also for the user intent

bayesian/api_v1.py
jyasveer: i want the user to send me both user and ecosystem

fridex: > i want the user to send me both user and ecosystem

OK, so in that case, I guess something like:

```python
if not user or not ecosystem:
```

should do the trick (empty strings). But I wonder if such can occur on flask level (not sure about it).

bayesian/api_v1.py
```diff
@@ -416,6 +417,56 @@ def post():
...
+            raise HTTPError(400, error="Expected user name in the request")
+
+        if ecosystem is None:
+            raise HTTPError(400, error="Expected ecosystem in the request")
+
+        s3 = StoragePool.get_connected_storage('S3ManualTagging')
+        # get user data
+        try:
+            result = s3.fetch_user_data(user, ecosystem)
+        except Exception as exc:
```
fridex: Please catch only `botocore.exceptions.ClientError` exception. This way we will return 404 every time that is not something we would like to do - this error can be confusing. Leave 500 in cases when something unpredictable happened (e.g. S3 is down).

```python
import botocore

...

except botocore.exceptions.ClientError:
    current_app.logger.exception("Failed to fetch data for the user {u}".format(u=user))
    raise HTTPError(404, error="Failed to fetch data for the user {u}, ecosystem {e}".format(u=user, e=ecosystem))
```

fridex: I've updated error hanlding.

jyasveer: this makes sense

miteshvp:
@fridex - can you merge this if you're ok with fixes now?

fridex:
> @fridex - can you merge this if you're ok with fixes now?

It requires https://github.com/fabric8-analytics/fabric8-analytics-worker/pull/402/ to go in first and bump worker image not to break anything.

README.md
```diff
@@ -80,4 +80,4 @@ The marked offline tests can then be run as:
 If the Docker container based tests have been run, then this may complain
 about mismatched locations in compiled files. Those can be deleted using:
 
-    (bayesian) $ sudo find -name *.pyc -delete
+    (bayesian) $ sudo find -name *.pyc -delete
```
fridex: Why was there added space?

requirements.txt
```diff
@@ -34,6 +34,7 @@ pytz==2016.7              # via flask-restful
 requests-futures==0.9.7
 requests==2.18.1          # via requests-futures
 six==1.10.0               # via flask-appconfig, flask-cors, flask-restful, python-dateutil
+selinon==0.1.0rc5
```
fridex: I'm not sure if this is needed as selinon is a direct dependency of worker so we don't need to state it twice. If updates will be issued, we will need to do changes twice with this.

(BTW next time you would like to introduce a dependency, place it to `requirements.in` and run `pip-compile` so deps are locked as expected).

bayesian/api_v1.py
fridex: Please respect review comments.

requirements.txt
jyasveer: i added this dependency here because i need the StoragePool object to get the s3 bucket details.

fridex: Did you experienced with some issues? This should be [already installed](https://github.com/fabric8-analytics/fabric8-analytics-server/blob/master/Dockerfile#L36) in the server image as a dependency of worker.

samuzzal-choudhury: @jyasveer, as @fridex mentioned, selinon is a direct dependency of worker and worker is a dependency for server. So, the server container will have worker and all it's dependencies installed and you should be able to use StoragePool object directly.

bayesian/api_v1.py
```diff
@@ -418,6 +420,60 @@ def post():
...
+            if 'user' not in input_json:
+                raise HTTPError(400, error="Expected user name in the request")
+
+            if 'data' not in input_json:
+                raise HTTPError(400, error="Expected tags in the request")
+
+            s3 = StoragePool.get_connected_storage('S3ManualTagging')
+
+            # Store data
+            return s3.store_user_data(input_json)
```
fridex: Ideally we would like to return something like 201 and empty dict or so, but OK for now.

PR #168, opened 2017-09-28
Allow URL encoded package names

tuxdna:
For: https://github.com/openshiftio/openshift.io/issues/686


tuxdna:
This is to enable us to pass package names containing slashes. 

Therefore the API request for component analysis of `github.com/golang/dep` version `v0.3.1` becomes:

```
http://localhost:32000/api/v1/component-analyses/go/github.com%2Fgolang%2Fdep/v0.3.1
```

fridex:
Maybe we would like to introduce something similar to our `ComponentSearch` endpoint, but that is another story...

PR #175, opened 2017-10-06
introduce the manifest generation logic

### LINKED ISSUE #174: [POC] Introduce the manifest generation functionality inside fabric8-server - Machine Generated Stacks
Description about machine generated stacks

https://github.com/openshiftio/openshift.io/issues/1005
https://github.com/openshiftio/openshift.io/issues/1037

ravsa:
1. Changes the api_v1.py
2. Introduces a new file generate_manifest.py

Fixes #174.

### DIFF
--- a/bayesian/api_v1.py
+++ b/bayesian/api_v1.py
@@ -595,7 +595,7 @@ class GenerateManifest(Resource):
     def post():
         input_json = request.get_json()
         if 'ecosystem' not in input_json:
-            raise HTTPError(400, "Must provide a ecosystem")
+            raise HTTPError(400, "Must provide an ecosystem")
         if input_json.get('ecosystem') == 'maven':
             return Response(
                 PomXMLTemplate(input_json).xml_string(),
--- a/bayesian/api_v1.py
+++ b/bayesian/api_v1.py
@@ -598,20 +598,18 @@ class GenerateManifest(Resource):
             raise HTTPError(400, "Must provide a ecosystem")
         if input_json.get('ecosystem') == 'maven':
             return Response(
-                PomXMLTemplate(input_json).xml_file(),
+                PomXMLTemplate(input_json).xml_string(),
                 headers={
                     "Content-disposition": 'attachment;filename=pom.xml',
                     "Content-Type": "text/xml;charset=utf-8"
                 }
             )
-        elif input_json.get('ecosystem') == 'node':
-            return {'result': 'ecosystem node is not yet supported'}
-        elif input_json.get('ecosystem') == 'python':
-            return {'result': 'ecosystem python is not yet supported'}
         else:
-            current_app.logger.exception('Unknown ecosystem {esystem} encountered'.format(
-                esystem=input_json.get('ecosystem')))
-            raise HTTPError(400, "Unknown ecosystem encountered")
+            return Response(
+                {'result': "ecosystem '{}' is not yet supported".format(
+                    input_json['ecosystem'])},
+                status=400
+            )
 
 
 add_resource_no_matter_slashes(ApiEndpoints, '')
--- a/bayesian/api_v1.py
+++ b/bayesian/api_v1.py
@@ -7,7 +7,7 @@ import re
 
 from io import StringIO
 
-from flask import Blueprint, current_app, request, url_for, send_file
+from flask import Blueprint, current_app, request, url_for, Response
 from flask.json import jsonify
 from flask_restful import Api, Resource, reqparse
 from flask_cors import CORS
@@ -96,15 +96,6 @@ def liveness():
     current_app.logger.warning("Liveness probe finished")
     return jsonify({}), 200
 
-
-@api_v1.route('/generatefile', methods=['POST'])
-def generatefile():
-# generate manifest file
-    input_json = request.get_json()
-    if input_json['ecosystem'] == 'maven':
-        return send_file(PomXMLTemplate(input_json).xml_file(),
-                        attachment_filename='pom.xml',
-                        as_attachment=True)
 
 api_v1.coreapi_http_error_handler = handle_http_error
 # work around https://github.com/flask-restful/flask-restful/issues/542
@@ -200,6 +191,7 @@ class ResourceWithSchema(Resource):
       on `GET` request.
     Note that if both `schema_ref` and `add_schema` are defined, only the method will be used.
     """
+
     def add_schema(self, response, status_code, method):
         """Add schema to response. The schema must be dict containing 3 string values:
         name, version and url (representing name and version of the schema and its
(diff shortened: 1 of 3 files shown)


bayesian/api_v1.py
```diff
@@ -96,6 +97,15 @@ def liveness():
     return jsonify({}), 200
 
 
+@api_v1.route('/generatefile', methods=['POST'])
```
fridex: To follow the current API naming schema, it would be better to name endpoint `/generate-file`.

bayesian/api_v1.py
```diff
@@ -96,6 +97,15 @@ def liveness():
...
+def generatefile():
+# generate manifest file
+    input_json = request.get_json()
+    if input_json['ecosystem'] == 'maven':
+        return send_file(PomXMLTemplate(input_json).xml_file(),
+                        attachment_filename='pom.xml',
+                        as_attachment=True)
```
fridex: Probably missing else branch? What is expected behaviour?

bayesian/generate_manifest.py
```diff
@@ -0,0 +1,75 @@
...
+from io import BytesIO
+
+
+class PomXMLTemplate:
+
+    def __init__(self, json_data=None):
+        if json_data:
+            self._data = dict(json_data)
+        else:
+            raise ValueError
```
fridex: A descriptive error should be present - pass what is wrong to exception constructor.

bayesian/generate_manifest.py
```diff
@@ -0,0 +1,75 @@
...
+
+from lxml.etree import SubElement, Element, ElementTree
+from io import BytesIO
+
+
+class PomXMLTemplate:
+
+    def __init__(self, json_data=None):
+        if json_data:
+            self._data = dict(json_data)
```
fridex: Why there is casting to dict? Also if you request `json_data` to be present do not assign default value in constructor.

bayesian/generate_manifest.py
```diff
@@ -0,0 +1,75 @@
...
+            raise ValueError
+        self.root = Element(
+            'project',
+            xmlns="http://maven.apache.org/POM/4.0.0",
+        )
+        self.tree = ElementTree(self.root)
+        self.create()
+
+    def create(self):
+        self._project = self._data.get('project', None)
```
fridex: The `get` method defaults to None if the given key is present, so no need to pass `None` (here and bellow as well):
```python
>>> {}.get('foo') is None
True
```

bayesian/generate_manifest.py
```diff
@@ -0,0 +1,75 @@
...
+            self.add_dependencies(self._data.get('dependencies', None))
+
+    def add_framework(self, fw):
+        frameworks = {'springboot': ["org.springframework.boot",
+                                     "spring-boot-starter-parent",
+                                     self._data.get('version', None)],
+                      'wildfly': (None, None, None),
+                      'vertx': (None, None, None)
+                      }
+        if fw in frameworks:
```
fridex: Probably some error when `fw` is not present?

bayesian/generate_manifest.py
```diff
@@ -0,0 +1,75 @@
...
+        if dependencies:
+            self.dpmanage = SubElement(self.root, "dependencyManagement")
+            self.dps = SubElement(self.dpmanage, "dependencies")
+            for item in dependencies:
+                dp = SubElement(self.dps, 'dependency')
+                for child, data in zip(('groupID', 'artifactID', 'version'),
+                                       item.split(':')):
+                    SubElement(dp, child).text = data
+
+    def xml_file(self):
```
fridex: Maybe it would be more readable to return call `tostring()` and assign correct content-type to response. WDYT?

bayesian/api_v1.py
```diff
@@ -96,6 +97,15 @@ def liveness():
     return jsonify({}), 200
 
 
+@api_v1.route('/generatefile', methods=['POST'])
+def generatefile():
```
fridex: Please create endpoint by deriving from `ResourceWithSchema` to utilize endpoint handling on API side.

ravsa: @fridex we don't want to expose this API end-point

ravsa: and I have made all the necessary changes in the code.

fridex: > @fridex we don't want to expose this API end-point

What do you mean by this? It needs to be exposed in order to access it.

ravsa: No, I mean API end-point not to be visible. like /liveness end-point.

fridex: > No, I mean API end-point not to be visible. like /liveness end-point.

I would say this corresponds to API endpoints which should be properly versioned and we should maintain compatibility between versions. `/liveness` and `/readiness` are here for openshift checks and they shouldn't be exposed outside cluster (they are now though). This endpoint will be exposed outside the cluster so users can reach it.

bayesian/api_v1.py
```diff
@@ -583,6 +585,32 @@ def get_component_analysis_schema_url(cls, name, version):
...
+            raise HTTPError(400, "Must provide a ecosystem")
+        if input_json.get('ecosystem') == 'maven':
+            return Response(
+                PomXMLTemplate(input_json).xml_file(),
+                headers={
+                    "Content-disposition": 'attachment;filename=pom.xml',
+                    "Content-Type": "text/xml;charset=utf-8"
+                }
+            )
+        elif input_json.get('ecosystem') == 'node':
```
fridex: What about having directly else branch so if we add a new ecosystem, the error message would be reusable (in other words do not hardcode ecosystem names in sources if not necessary)?
```python
else:
    return {'result': "ecosystem '{}' is not yet supported".format(input_json['ecosystem'])}
```

Also return 400 HTTP status code to signalize error instead of default 200.

bayesian/api_v1.py
```diff
@@ -583,6 +585,32 @@ def get_component_analysis_schema_url(cls, name, version):
...
+    method_decorators = [login_required]
+
+    @staticmethod
+    def post():
+        input_json = request.get_json()
+        if 'ecosystem' not in input_json:
+            raise HTTPError(400, "Must provide a ecosystem")
+        if input_json.get('ecosystem') == 'maven':
+            return Response(
+                PomXMLTemplate(input_json).xml_file(),
```
fridex: Method name `xml_file` is misleading. Probably something like xml_string() would be more accurate.

bayesian/api_v1.py
ravsa: yes, this makes sense. I will change this.

fridex:
[test]

fridex:
@ravsa please take a look at [failing test case](https://ci.centos.org/job/devtools-fabric8-analytics-server-fabric8-analytics/72/console). You need to add newly introduced endpoint to `TestApiV1Root.api_root` listing.

ravsa:
[test]

samuzzal-choudhury:
[test]

fridex:
@samuzzal-choudhury could you check this?

bayesian/api_v1.py
```diff
@@ -583,6 +585,30 @@ def get_component_analysis_schema_url(cls, name, version):
...
 
 
+class GenerateManifest(Resource):
+    method_decorators = [login_required]
+
+    @staticmethod
+    def post():
+        input_json = request.get_json()
+        if 'ecosystem' not in input_json:
+            raise HTTPError(400, "Must provide a ecosystem")
```
samuzzal-choudhury: typo 'a' has to be replaced with 'an'

samuzzal-choudhury:
[test]

PR #185, opened 2017-10-18
Added separate methods for post and get for user intent

jyasveer:


msrb:
unit tests are failing: https://ci.centos.org/job/devtools-fabric8-analytics-server-fabric8-analytics/86/console

bayesian/api_v1.py
```diff
@@ -679,6 +692,7 @@ def post():
 add_resource_no_matter_slashes(StackAnalysesGET, '/stack-analyses/<external_request_id>')
 add_resource_no_matter_slashes(UserFeedback, '/user-feedback')
 add_resource_no_matter_slashes(UserIntent, '/user-intent')
+add_resource_no_matter_slashes(UserIntentGet, '/user-intent/<user>/<ecosystem>')
```
fridex: What about adjusting only the line above? That should do the trick. Everything else can stay untouched.

jyasveer: how do you intent to do it ?

bayesian/api_v1.py
```diff
@@ -425,26 +425,11 @@ class UserIntent(ResourceWithSchema):
...
-                                                                                    e=ecosystem)
-            current_app.logger.exception(err_msg)
-            raise HTTPError(404, error=err_msg)
-
-        return result
+        raise HTTPError(404, "Unsupported API endpoint")
 
     @staticmethod
     def post():
+        decoded = decode_token()
```
msrb: Do we need this here?

bayesian/api_v1.py
```diff
@@ -425,26 +425,11 @@ class UserIntent(ResourceWithSchema):
...
-        try:
-            result = s3.fetch_user_data(user, ecosystem)
-        except botocore.exceptions.ClientError:
-            err_msg = "Failed to fetch data for the user {u}, ecosystem {e}".format(u=user,
-                                                                                    e=ecosystem)
-            current_app.logger.exception(err_msg)
-            raise HTTPError(404, error=err_msg)
-
-        return result
+        raise HTTPError(404, "Unsupported API endpoint")
```
msrb: 405 would be probably more appropriate here. But this should be transparently handled by flask (webserver? not sure). So we can remove the `get()` method altogether.

bayesian/api_v1.py
```diff
@@ -474,6 +459,34 @@ def post():
...
+            err_msg = "Failed to fetch data for the user {u}, ecosystem {e}".format(u=user,
+                                                                                    e=ecosystem)
+            current_app.logger.exception(err_msg)
+            raise HTTPError(404, error=err_msg)
+
+        return result
+
+    @staticmethod
+    def post():
+        raise HTTPError(404, "Unsupported API endpoint")
```
msrb: 405, or remove :) I'd prefer removing this method.

bayesian/api_v1.py
fridex: See https://github.com/fabric8-analytics/fabric8-analytics-server/pull/186 if that would work for you.

jyasveer: if i have a single end point , how will my post request get served ? @fridex

fridex: Agh, good point, I missed the POST endpoint. PR has been adjusted, please take a look.

PR #186, opened 2017-10-18
Fix user intent endpoint parameters

fridex:


fridex:
@jyasveer could you very that this fixes your issues? If yes, please close https://github.com/fabric8-analytics/fabric8-analytics-server/pull/185

PR #171, opened 2017-10-05
populate simple reasons for recommendation

SaketJNU:
This PR is opened to for the issue https://github.com/openshiftio/openshift.io/issues/984. Here, the objective is to populate reason in simple sentence for each recommended alternate  and companion package.

PR #188, opened 2017-10-19
Revert "populate simple reasons for recommendation"

samuzzal-choudhury:
Reverts fabric8-analytics/fabric8-analytics-server#171


miteshvp:
@msrb @samuzzal-choudhury - do we know the reason for its failure? Is it because of schema? I mean there were keys added to schema?

miteshvp:
@msrb @samuzzal-choudhury - is it ok to retry to push this code again?

PR #189, opened 2017-10-23
[openshift] More RAM and CPU

msrb:


msrb:
cc @samuzzal-choudhury

PR #190, opened 2017-10-23
Added the step to connect to celery if not connected

jyasveer:


jyasveer:
makes sense @fridex . will update it

jyasveer:
@fridex 
i looked at the code. i am doing what you referred in the other comment already. connect_if_not_connected() in setup.py is calling init_selinon() . please refer here https://github.com/fabric8-analytics/fabric8-analytics-server/blob/master/bayesian/setup.py#L24

fridex:
> @fridex
i looked at the code. i am doing what you referred in the other comment already. connect_if_not_connected() in setup.py is calling init_selinon() . please refer here https://github.com/fabric8-analytics/fabric8-analytics-server/blob/master/bayesian/setup.py#L24

Yes, that's true. But it also initializes connection to message broker that is AFAIKS not needed in these cases. Using `init_selinon` is more optimal.

PR #193, opened 2017-11-01
Improve logging and debugging

msrb:


msrb:
Related to https://github.com/openshiftio/openshift.io/issues/1243.

@tisnik I'd try something like this first. WDYT?

cc @samuzzal-choudhury

tisnik:
LGTM! Thanks a lot

PR #204, opened 2017-11-13
Crowd sourcing API Implementation for getting component to be tagged and setting tags to component.

ravsa:
1. Get next untagged component for an ecosystem
    - Component from the same ecosystem that user is familiar to (based on IDE or user choosing)
    - The Same user should not tag the same component twice

2. Set the tags when the user submits them
    - Create a user if does not exist
    - Create an edge if not exists such as User -> has_tagged -> Package


bayesian/utils.py
```diff
@@ -337,9 +342,13 @@ def build_nested_schema_dict(schema_dict):
 
 def get_next_component_from_graph(ecosystem, user_id):
     tag_count = 2
-    qstring = "user = g.V().has('userid', '{uid}').next();\
-            g.V(user).as('u').V().has('ecosystem','{e}').has('manual_tagging_required', true).has('tags_count', not(within({tc}))).coalesce(inE('has_tagged').as('pkg').outV().as('b').where('u', neq('b')).outE('has_tagged').inV().as('np').where('pkg', neq('np')), V().has('ecosystem','{e}').has('manual_tagging_required', true).not(inE('has_tagged'))).limit(1).valueMap();".format(
-                    uid=user_id, e=ecosystem, tc=tag_count)
+    qstring = ("user = g.V().has('userid', '{uid}').next();"
```
miteshvp: @ravsa - on second thoughts, I guess, its better to create a user node while fetching. You can implement the way it is done below
`"user = g.V().has('userid','{user_id}').tryNext().orElseGet{{graph.addVertex("
               "'vertex_label', 'User', 'userid', '{user_id}', {company})}};
`

bayesian/utils.py
```diff
@@ -340,15 +340,21 @@ def build_nested_schema_dict(schema_dict):
     return result
 
 
-def get_next_component_from_graph(ecosystem, user_id):
+def get_next_component_from_graph(ecosystem, user_id, company):
```
miteshvp: maybe `company=None`, because you might not get company decoded if its anoffline token or minimalist token

bayesian/utils.py
```diff
@@ -340,15 +340,21 @@ def build_nested_schema_dict(schema_dict):
...
     tag_count = 2
-    qstring = ("user = g.V().has('userid', '{uid}').next();"
+    qstring = ("user = g.V().has('userid','{uid}').tryNext().orElseGet{{graph.addVertex("
+               "'vertex_label', 'User', 'userid', '{uid}', {company})}};"
```
miteshvp: "'vertex_label', 'User', 'userid', '{uid}', {company})}};"
Extra comma after '{uid}' needs to be removed

bayesian/utils.py
```diff
@@ -339,12 +340,82 @@ def build_nested_schema_dict(schema_dict):
     return result
 
 
+def get_next_component_from_graph(ecosystem, user_id, company):
+    tag_count = 2
```
samuzzal-choudhury: @ravsa this tag count should be provided as a configurable in `default_config.py` and it should be made available through an environment variable.

You can make an entry for a new environment variable in openshift/template.yaml.

PR #210, opened 2017-11-16
Added home area for user coreapi

samuzzal-choudhury:
This should fix the issue https://github.com/openshiftio/openshift.io/issues/814.

The theory is that - coreapi user is created without a home area whereas .gitconfig file needs a home to get created.


msrb:
Hmm... wait, so we are cloning git repositories in the context of a request? That doesn't seem right :/ LGTM for this change, but I suspect there will be more problems...

PR #213, opened 2017-11-20
Fix Pylint warnings in Alembic directory

tisnik:
We are going to check ALL sources (even the generated ones), so it is needed to fix them all.


pkajaba:
It looks good, but I have just one comment. Why don't you enable checking of alembic dir in this commit?

pkajaba:
@tisnik ^^

tisnik:
@pkajaba I need to fix other sources as well. I'm going to prepare bigger PR for tomorrow (I'm just in the middle of work) and it definitely will contain update for the Pylinter script as well.

PR #214, opened 2017-11-21
Fix all remainded alembic warnings

tisnik:
Now all source files are checked by Pylint:
https://fabric8-analytics.github.io/dashboard/dashboard.html

Also detailed report:
https://fabric8-analytics.github.io/dashboard/repository_fabric8-analytics-server.html


check-docstyle.sh
```diff
@@ -1,4 +1,4 @@
-directories="bayesian hack tests"
+directories="bayesian hack tests alembic"
```
pkajaba: why don't you include here also setup.py?

PR #217, opened 2017-11-29
[WIP] Add endpoints and implementation for bookkeeping API

tuxdna:
Fixes issue:

 * https://github.com/openshiftio/openshift.io/issues/1404


msrb:
I am wondering if `jobs` wouldn't be a better place where to implement this API. I know `jobs` is quite overloaded already with different API endpoints, and we should clean it up some day, but it's currently the API for managing the data ingestion part of the system. This API server is an entry point for OSIO users and they probably won't be interested in our internal bookkeeping data.

@tuxdna what do you think Saleem?

tuxdna:
@msrb: I was also thinking about putting this logic in Jobs API as well. Yes we can instead move bookkeeping to Jobs.

tuxdna:
Closing because this API will be part of Jobs now - https://github.com/fabric8-analytics/fabric8-analytics-jobs/pull/218

PR #220, opened 2017-12-18
Add companion confidence and count statements

sara-02:
Part of user-story https://github.com/openshiftio/openshift.io/issues/1674

### DIFF
--- a/bayesian/utils.py
+++ b/bayesian/utils.py
@@ -31,6 +31,9 @@ gremlin_url = "http://{host}:{port}".format(
     host=os.environ.get("BAYESIAN_GREMLIN_HTTP_SERVICE_HOST", "localhost"),
     port=os.environ.get("BAYESIAN_GREMLIN_HTTP_SERVICE_PORT", "8182"))
 
+companion_reason_statement = "along with the provided input stack. " \
+    "Do you want to consider adding this Package?"
+
 
 def get_recent_analyses(limit=100):
     return rdb.session.query(Analysis).order_by(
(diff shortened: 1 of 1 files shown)


bayesian/utils.py
```diff
@@ -624,15 +627,17 @@ def _check_usage_outlier(self, pkg_name, manifest_response):
 
     def _companion_reason(self, manifest_response):
         """
-        It will populate the simple reason for each companion package
+        It will populate the simple reason for each companion package.
```
pkajaba: please follow https://www.python.org/dev/peps/pep-0257/ so we don't have to manually do everything.

sara-02: @pkajaba I did not write the doc-string, just updated it as per our `doc-string checker`.

sara-02:
@pkajaba I have added the user-story in the description section, should suffice your questions.

sara-02:
@miteshvp we will have confidence score as column, and hover-reasoning part can remain same as it was before.

bayesian/utils.py
pkajaba: yes, but if you change ones which you encounter (which is fairly easy) we end up with better coverage and less work in future.

pkajaba:
>  I have added the user-story in the description section, should suffice your questions.

@sara-02 I see, thanks. However, I think that this kind of reasoning should be generated at the client side. @miteshvp what do you think?

bayesian/utils.py
```diff
@@ -624,15 +627,17 @@ def _check_usage_outlier(self, pkg_name, manifest_response):
...
             name = pkg.get("name")
-            stack_count = str(pkg.get("cooccurrence_probability"))
-            sentence = "Package {} appears in {} different stacks along with the provided input " \
-                       "stack. Do you want to consider adding this Package?"\
-                .format(name, stack_count)
-            pkg["reason"] = sentence
+            stack_confidence = str(pkg.get("cooccurrence_probability"))
+            stack_count = str(pkg.get("cooccurrence_count"))
+            count_sentence = "Package {} appears in {} different stacks".format(
+                name, stack_count)
```
tisnik: I would not add the whole sentences into the REST API response. This is job for a consumer to do it properly (with possible translations etc. etc.)

miteshvp: @tisnik - IMHO, client should be dumb enough to populate key value pairs. Having said that, we maintain different UI (OSIO stack report, stack-report portal, VSCode IDE, Devstudio IDE) and in the future we will add more. Hence it made more sense to send the reasoning from API server so that all clients gets exact messaging as we need plus lot less maintenance. I hope I answered your query.

pkajaba: @miteshvp OK, so @sara-02  would you kindly update documentation of our API endpoint to reflect this new field?

sara-02: @pkajaba please share the link for the same.

pkajaba: cc @samuzzal-choudhury I can't find it, but there should be anything like that, right?

sara-02: @pkajaba does not look like it is in this repo, will have to create a new PR in which ever repo it is.

miteshvp:
LGTM

PR #223, opened 2017-12-21
Fixed some docstrings

tisnik:


bayesian/auth.py
```diff
@@ -45,6 +46,7 @@ def decode_token():
 
 
 def login_required(view):
+    """Check if the login is required and if the used can be authorized."""
```
jpopelka: used -> user

hack/auth_test_fixtures/generate_auth_token.py
```diff
@@ -1,7 +1,7 @@
-'''
-Script to generate an authentication token to be passed as header
-with requests to Bayesian API
-'''
+"""Script to generate an authentication token.
+
+This tokes is to be passed as header with requests to Bayesian API.
```
jpopelka: tokes -> token ?

PR #230, opened 2018-01-07
Introducing high performant endpoint analysis to perform stack analysis

samuzzal-choudhury:


bayesian/api_v1.py
```diff
@@ -840,7 +845,116 @@ def post():
...
+            api_url = current_app.config['F8_API_BACKBONE_HOST']
+
+            d = DependencyFinder()
+            deps = d.execute(args, rdb.session)
+            deps['external_request_id'] = request_id
+
+            _session.post('{}/api/v1/stack_aggregator'.format(api_url), json=deps)
+            _session.post('{}/api/v1/recommender'.format(api_url), json=deps)
+        except Exception as exc:
+            print(exc)
```
miteshvp: remove print debugs

hack/coreapi-server.sh
```diff
@@ -12,13 +12,15 @@ exec mod_wsgi-express start-server \
                       --port 5000 \
                       --user coreapi \
                       --group coreapi \
-                      --processes 4 \
-                      --threads 10 \
+                      --processes 1 \
```
miteshvp: Let's try load testing with multi-process multi-threads here.

samuzzal-choudhury: @miteshvp need to try and check. Mercator CLI runs on a thread and having a thread to run another thread basically halts the processing and results in gateway timeout. I will do my research and get back on this.

samuzzal-choudhury: @miteshvp in order to run stack analysis with the verified performance results, we need this configuration in coreapi-server.sh

miteshvp:
LGTM.

PR #231, opened 2018-01-08
Add missing backbone server port in environment variable

samuzzal-choudhury:
The default value for `F8_API_BACKBONE_HOST` environment variable should be set to `"http://f8a-server-backbone:5000"`


miteshvp:
LGTM

PR #243, opened 2018-01-31
functionality to store dependency snapshot in database

ravsa:
- store dependency snapshot in database or update existing snapshot
    `/api/v1/stack-analyses?persist=<true|false>&requestId=<request_id>`
- minor bug fixes

### DIFF
--- a/bayesian/api_v1.py
+++ b/bayesian/api_v1.py
@@ -17,6 +17,7 @@ from flask_restful import Api, Resource, reqparse
 from flask_cors import CORS
 from sqlalchemy import or_
 from sqlalchemy.exc import SQLAlchemyError
+from sqlalchemy.dialects.postgresql import insert
 from selinon import StoragePool
 
 from f8a_worker.models import (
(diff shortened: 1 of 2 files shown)


humaton:
[test]

bayesian/api_v1.py
```diff
@@ -919,35 +924,53 @@ def post():
...
             _session.post('{}/api/v1/stack_aggregator'.format(api_url), json=deps)
             _session.post('{}/api/v1/recommender'.format(api_url), json=deps)
         except Exception as exc:
             raise HTTPError(500, ("Could not process {t}.".format(t=request_id))) from exc
-
-        return {"status": "success", "submitted_at": str(dt), "id": str(request_id)}
+        if persist:
+            if sid:
+                try:
+                    result = rdb.session.query(StackAnalysisRequest)\
```
miteshvp: You can also look at http://docs.sqlalchemy.org/en/latest/dialects/postgresql.html?highlight=conflict#insert-on-conflict-upsert

miteshvp:
@ravsa - Let's discuss this workflow again. Although it looks good, but please change this PR in WIP till then

bayesian/api_v1.py
ravsa: Thanks @miteshvp, I will do the changes.

ravsa:
sure, current workflow 
1) If no parameters passed API will work as it was working before.
2) If `persist=true`  and `sid` is passed then it will update the dep-snapshot.
3) if `persist=false` is passed as a parameter then no insertion and update in the database.

miteshvp:
> If persist=true and sid is passed then it will update the dep-snapshot.

We just need to upsert in API server without any parameters actually needed (based on sid present or not).
`persist=false` will help backbone to take a decision whether to persist the record or not

PR #252, opened 2018-02-02
lets move to http probe

miteshvp:


miteshvp:
merging myself as these are template changes

PR #254, opened 2018-02-02
move to tcp probe again

miteshvp:


miteshvp:
merged it myself. only config changes

PR #255, opened 2018-02-02
Documentation fixes

tisnik:


MAINTAINERS
```diff
@@ -1,3 +1,6 @@
 Bohuslav Kabrda <bkabrda@redhat.com>
 Pavel Odvody <podvody@redhat.com>
 Tomáš Tomeček <ttomecek@redhat.com>
+Samuzzal Choudhury
+Michal Srb <michal@redhat.com>
+Pavel Tišnovský <ptisnovs@redhat.com>
```
jpopelka: Sam has no e-mail ? ;-)
I think we can at the same time remove the old maintainers, since non of them is maintaining it any more.

PR #264, opened 2018-02-22
Set Access-Control-Allow-Origin for stack analyses

samuzzal-choudhury:


miteshvp:
Merging to check a fix for a defect

PR #267, opened 2018-02-22
Added some more relevant headers and removed CORS module

samuzzal-choudhury:


jpopelka:
> But linter is complaining

Feel free to ignore that, somebody else will [fix it](
b2cec4c)

PR #276, opened 2018-02-27
Service to detect under what license a project is distributed

ravsa:
License extractor scans the LICENSE files and compare the contents with known license synonyms and detects the license for the project

### DIFF
--- a/bayesian/default_config.py
+++ b/bayesian/default_config.py
@@ -24,6 +24,13 @@ SECRET_KEY = 'devkey'
 # length of token life in seconds
 API_TOKEN_LIFETIME = 3600
 
+# license synonyms
+# /<url>/<org>/<repo>/<branch>/<dir/file>
+LIC_SYNONYMS_URL = '/'.join(['https://raw.githubusercontent.com', 'fabric8-analytics',
+                             'fabric8-analytics-license-analysis', 'master',
+                             'tests/synonyms/license_synonyms.json'])
+
+
 # Info about deployed version
 SYSTEM_VERSION = os.getenv('F8A_SYSTEM_VERSION', default='/etc/coreapi-release')
 
--- a/bayesian/license_extractor.py
+++ b/bayesian/license_extractor.py
@@ -1,0 +1,56 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+"""License extractor utility functions."""
+
+from .default_config import LIC_SYNONYMS_URL
+from requests import get
+from collections import defaultdict
+from lru import lru_cache_function
+from flask import current_app
+from datetime import datetime as dt
+
+
+# caching response for 24 hours
+@lru_cache_function(max_size=2048, expiration=60 * 60 * 24)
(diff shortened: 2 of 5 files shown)


bayesian/license_extractor.py
```diff
@@ -0,0 +1,56 @@
...
+    """Fetch all the license sysnonyms from license anlysis github repo."""
+    resp = get(LIC_SYNONYMS_URL)
+    if resp.status_code == 200:
+        current_app.logger.info(
+            "{} Succefully fetched license synonyms".format(dt.now()))
+        return resp.json()
+    else:
+        current_app.logger.error("{tm} Unable to fetch license synonyms, STATUS_CODE:{cd}".format(
+            tm=dt.now(), cd=resp.status_code))
+        return {}
```
msrb: Returning empty dictionary here means that it will stay cached for the whole day. Another possible approach could be to throw an exception here and thus skip caching when the GitHub request above failed from some reason.

bayesian/default_config.py
```diff
@@ -24,6 +24,13 @@
 # length of token life in seconds
 API_TOKEN_LIFETIME = 3600
 
+# license synonyms
+# /<url>/<org>/<repo>/<branch>/<dir/file>
+LIC_SYNONYMS_URL = '/'.join(['https://raw.githubusercontent.com', 'fabric8-analytics',
```
msrb: I realize that this is very convenient, but it also can make debugging harder as nobody will be sure what version exactly was used by the service when a problem occurred. Just saying :)

bayesian/license_extractor.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def extract_licenses(license_files):
```
msrb: This function deserves some comments ;)

bayesian/default_config.py
ravsa: yes, I agree with you but this is just a JSON file contains synonyms list if it has some problem it will occur in `fabric8-analytics-license-analysis` first.

bayesian/license_extractor.py
ravsa: Thanks, for pointing that out I will do the necessary changes.

bayesian/license_extractor.py
ravsa: ok, I'll do that :)

bayesian/license_extractor.py
```diff
@@ -0,0 +1,63 @@
...
+    """Extract license name from the given license files."""
+    lic_syn = get_license_synonyms()
+    response = dict()
+    if lic_syn:
+        for f_no, _file in enumerate(license_files, 1):
+            license_key = None
+            try:
+                # remove all the unnecessary chars ',', \n, space
+                content = _file.read().decode('utf-8')
```
jpopelka: In Python 3 `read()` returns `str`, i.e. decoded string, so the `decode('utf-8')` won't actually work

miteshvp:
Merging this since all comments are handled.

PR #284, opened 2018-03-09
Next Best Alternate version if input version has a CVE

miteshvp:
If an EPV given by a user contains a CVE then give user the next higher version with no CVE

### DIFF
--- a/bayesian/utils.py
+++ b/bayesian/utils.py
@@ -204,7 +204,7 @@ def generate_recommendation(data, package, input_version):
             graph_ver = convert_version_to_proper_semantic(graph_version)
 
             # Check for next best higher version than input version without any CVE's
-            if (ver.get('cve_ids', [''])[0] != '' or 'cve_ids' not in ver) \
+            if (ver.get('cve_ids', [''])[0] == '' or 'cve_ids' not in ver) \
                     and sv.Version(graph_ver) > sv.Version(ip_ver):
                 if not higher_version:
                     higher_version = graph_ver
--- a/bayesian/utils.py
+++ b/bayesian/utils.py
@@ -204,8 +204,7 @@ def generate_recommendation(data, package, input_version):
             graph_ver = convert_version_to_proper_semantic(graph_version)
 
             # Check for next best higher version than input version without any CVE's
-            if (ver.get('cve_ids', [''])[0] == '' or 'cve_ids' not in ver) \
-                    and sv.Version(graph_ver) > sv.Version(ip_ver):
+            if not ver.get('cve_ids') and sv.Version(graph_ver) > sv.Version(ip_ver):
                 if not higher_version:
                     higher_version = graph_ver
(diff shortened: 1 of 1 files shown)


bayesian/utils.py
```diff
@@ -198,23 +198,22 @@ def generate_recommendation(data, package, version):
...
-                        if cvss >= max_cvss:
-                            break
-                message += '\n It is recommended to use Version - ' + latest_version
-                reco['recommendation']['change_to'] = latest_version
-                reco['recommendation']['message'] = message
+            graph_version = ver.get('version', [''])[0]
+            graph_ver = convert_version_to_proper_semantic(graph_version)
+
+            # Check for next best higher version than input version without any CVE's
+            if (ver.get('cve_ids', [''])[0] != '' or 'cve_ids' not in ver) \
```
jpopelka: Hopefully this can wait till Monday, because my tired brain is failing to process this.

Looks like `'if (cve_ids is there and not empty) or it's not there'`, which I don't understand.

tisnik: it would help to have such code covered by unit tests ;) is it possible to spend some time on them?

miteshvp: You are right. I wanted `if (cve_ids is there and empty) or it's not there`. Fixed it

miteshvp: @tisnik - I fixed the if condition but what I was trying with this condition is to make it future change prone if we make our graphDB consistent (create properties will null values). I don't think we require a unit test here.

jpopelka: Ok, then I think we can simplify this to `if not ver.get('cve_ids'):` which should do the same, i.e. be true if the `'cve_ids'` is not there or is empty (or None).

bayesian/utils.py
```diff
@@ -198,23 +198,22 @@ def generate_recommendation(data, package, version):
...
+            graph_version = ver.get('version', [''])[0]
+            graph_ver = convert_version_to_proper_semantic(graph_version)
+
+            # Check for next best higher version than input version without any CVE's
+            if (ver.get('cve_ids', [''])[0] == '' or 'cve_ids' not in ver) \
+                    and sv.Version(graph_ver) > sv.Version(ip_ver):
+                if not higher_version:
+                    higher_version = graph_ver
+                if sv.Version(higher_version) > sv.Version(graph_ver):
+                    higher_version = graph_ver
```
jpopelka: Shouldn't this be:
```python
if sv.Version(graph_ver) > sv.Version(higher_version):
    higher_version = graph_ver

```

miteshvp: Actually we want the closest upgraded version to a user's version. If user's version is 1.0.0 and we have 1.0.1 and 1.0.4 without any CVE, we would like to recommend a closer version to the user's which would be 1.0.1 in this case.

jpopelka: makes sense :+1:

PR #286, opened 2018-03-14
[WIP] Repr_license_finder for go

vasulakkaraju:
Module added for fetching repr license for go by considering all dependencies(direct and transitive deps) .do not merge or review currently.


bayesian/api_v1.py
```diff
@@ -111,6 +114,17 @@ def liveness():
     return jsonify({}), 200
 
 
+@api_v1.route('/repr_licesne')
```
ravsa: I think you should implement this API end-point something like this https://github.com/fabric8-analytics/fabric8-analytics-server/blob/master/bayesian/api_v1.py#L490

bayesian/api_v1.py
```diff
@@ -111,6 +114,17 @@ def liveness():
...
+def test_repr_license():
+    """Handle the POST REST API call."""
+    args = get.input_json()
+    if args['ecosystem'] == 'go':
```
ravsa: It would be better if you use args.get('ecosystem'), think about a condition where key `ecosystem` doesn't exist it will throw `KeyError` which you are not handling here.

bayesian/api_v1.py
msrb: +1, please make use of flask-restfull when we already use it in the code.

bayesian/api_v1.py
```diff
@@ -111,6 +114,17 @@ def liveness():
...
+        rep_license = get_repr_license(args['ecosystem'], args['url'], LICENSE_SCORING_URL_REST)
```
msrb: The same as above, we should probably first validate the input.

bayesian/api_v1.py
```diff
@@ -111,6 +114,17 @@ def liveness():
...
+        return rep_license.caller()
+    else:
+        print("Currently only go lang repositories are supported!!")
```
msrb: Please use Flask logger, i.e.: `current_app.logger.info('message')`

msrb: Also we should probably bail out here, if the ecosystem is not supported.

bayesian/utils.py
```diff
@@ -32,9 +32,13 @@
...
-gremlin_url = "http://{host}:{port}".format(
-    host=os.environ.get("BAYESIAN_GREMLIN_HTTP_SERVICE_HOST", "localhost"),
-    port=os.environ.get("BAYESIAN_GREMLIN_HTTP_SERVICE_PORT", "8182"))
+GREMLIN_SERVER_URL_REST = "http://{host}:{port}".format(
+    host=os.environ.get("BAYESIAN_GREMLIN_HTTP_SERVICE_HOST"),
+    port=os.environ.get("BAYESIAN_GREMLIN_HTTP_SERVICE_PORT"))
+
+LICENSE_SCORING_URL_REST = "http://{host}:{port}".format(
+    host=os.environ.get("LICENSE_SERVICE_HOST"),
+    port=os.environ.get("LICENSE_SERVICE_PORT"))
```
msrb: These environment variables are not defined anywhere (?) 

You could use: `F8A_LICENSE_ANALYSIS_SERVICE_HOST` and `F8A_LICENSE_ANALYSIS_SERVICE_PORT`. Those two will be provided automatically by OpenShift.

vasulakkaraju: These two variables (LICENSE_SERVICE_HOST,LICENSE_SERVICE_PORT) are defined as environment variables in devcluster

bayesian/api_v1.py
```diff
@@ -1185,6 +1189,33 @@ def get(runtime):
...
+        if 'ecosystem' not in input_json:
+                raise HTTPError(400, error="Expected ecosystem in the request")
+
+        if input_json.get('ecosystem') == 'go':
+            rep_license = GetReprLicense(input_json['url'],
+                                         input_json['ecosystem'],
+                                         LICENSE_SCORING_URL_REST)
+            return rep_license.caller()
+        else:
+            current_app.logger.info("Currently only go lang repositories are supported!!")
```
tuxdna: Must always return a result. You might want to include info in returned JSON.

bayesian/repr_license_finder_go.py
```diff
@@ -0,0 +1,114 @@
...
+                                                "licenses": []})
+                    dep_pkg_list_unknown.append(self.ecosystem + ":" + name + ":" + version)
+                    continue
+
+        return {"packages": license_dict_modify}
+
+    def caller(self):
+        """Execute method for all the above functions."""
+        try:
+            r = requests.post(self.license_api, json=self.fetch_license_from_graph(
```
tuxdna: You may want to break this into multiple steps for clarity:


```python
glide_data = fetch_file_from_github(self.url, 'glide.lock')
extracted_info = self.godep_extractor(glide_data)
license_json=self.fetch_license_from_graph(**extracted_info)
r = requests.post(self.license_api, json=license_json)
```

If something causes in exception in above code, we will know exact line number which caused the error.

bayesian/repr_license_finder_go.py
```diff
@@ -0,0 +1,114 @@
...
+from .utils import fetch_file_from_github
+import requests
+import logging
+import traceback
+
+
+class GetReprLicense():
+    """Summary.
+
+    input:url:repo url to find repr_license
```
tuxdna: This documentation should be moved to `__init__()` instead. Please summarize the overall work done by GetReprLicense here.

bayesian/repr_license_finder_go.py
```diff
@@ -0,0 +1,114 @@
...
+    output:repr_license.
+    """
+
+    def __init__(self, url, ecosystem, license_api):
+        """Get all the inputs required to the class."""
+        self.url = url
+        self.ecosystem = ecosystem
+        self.license_api = license_api
+
+    def fetch_go_deps_from_github(self):
```
ravsa: @vasulakkaraju you can use the `fetch_file_from_github` function from `utils` I don't think you need to define this function again.

vasulakkaraju: yes the same function is used anyhow ill remove this definition from here @ravsa Thanks for your comment.

tisnik:
@miteshvp are you planning to merge this PR in the future or should it be closed?

miteshvp:
Closing this as an extension of this is available at #288

PR #288, opened 2018-03-18
added repr_license

vasulakkaraju:


bayesian/repr_license_finder.py
```diff
@@ -0,0 +1,88 @@
...
+from . import rdb, cache
+from .utils import fetch_file_from_github
+import requests
+import logging
+import traceback
+
+
+# Returns glide.lock file content as  a string for the github url given
+
+class GetReprLicense(object):
```
msrb: Docstrings are missing in this class. That seems to be the reason why CI is not happy:

```
bayesian/repr_license_finder.py
bayesian/repr_license_finder.py:1 at module level:
        D100: Missing docstring in public module
bayesian/repr_license_finder.py:19 in public class `GetReprLicense`:
        D101: Missing docstring in public class
bayesian/repr_license_finder.py:21 in public method `__init__`:
        D107: Missing docstring in __init__
bayesian/repr_license_finder.py:27 in public method `godep_extractor`:
        D102: Missing docstring in public method
bayesian/repr_license_finder.py:38 in public method `fetch_license_from_graph`:
        D102: Missing docstring in public method
bayesian/repr_license_finder.py:73 in public method `caller`:
        D102: Missing docstring in public method
6
    Fail
```

vasulakkaraju: @msrb earlier build was failing due to some other issues which is fixed now.Anyhow reg docstrings i am yet to add docstrings and fix it.Thanks !

tuxdna:
LGTM. Also remove `WIP` from the title.

miteshvp:
@tisnik - there are still certain items pending. I see we have covered quite a bit of ground here. But for now, we can close this and take the necessary review items when we reconsider this use-case. Thanks.

PR #296, opened 2018-04-05
Add parameter "source" to stack-analyses

samuzzal-choudhury:


bayesian/dependency_finder.py
```diff
@@ -46,6 +48,14 @@ def execute(self, arguments, db, manifests):
         # If we receive a manifest file we need to save it first
         result = []
         for manifest in manifests:
+            resolve_poms = False
+            if source == 'osio':
+                content_hash = generate_content_hash(manifest['content'])
+                if pom_hash.get(content_hash) is not None:
+                    return {'result': pom_hash.get(content_hash)}
```
miteshvp: this early return will not work for multiple manifests

bayesian/dependency_finder.py
```diff
@@ -104,4 +114,7 @@ def _flatten(deps, collect):
                 out["details"][0]['_resolved'] = resolved_deps
             result.append(out)
 
+        if source == 'osio':
```
miteshvp: storing of contents should be per manifest. This loop should be part of main for loop and not outside.

bayesian/dependency_finder.py
```diff
@@ -104,4 +114,7 @@ def _flatten(deps, collect):
...
+            pom_hash[content_hash] = result
```
miteshvp: pom_hash[content_hash] = out

bayesian/dependency_finder.py
samuzzal-choudhury: good catch

samuzzal-choudhury:
[test]

bayesian/dependency_finder.py
```diff
@@ -46,6 +49,15 @@ def execute(self, arguments, db, manifests):
...
         result = []
         for manifest in manifests:
+            content_hash = None
+            if source == 'osio':
+                content_hash = generate_content_hash(manifest['content'])
+                print("{} file digest is {}".format(manifest['filename'], content_hash))
+
+                s3 = AmazonS3(bucket_name='boosters-manifest')
+                s3.connect()
+                manifest['content'] = s3.retrieve_blob(content_hash).decode('utf-8')
```
miteshvp: we should handle error scenario as well

tisnik:
@samuzzal-choudhury could you please provide unit tests for the added/changed code? TY in advance.

bayesian/dependency_finder.py
```diff
@@ -46,6 +50,18 @@ def execute(self, arguments, db, manifests):
         # If we receive a manifest file we need to save it first
         result = []
         for manifest in manifests:
+            content_hash = None
+            if source == 'osio':
+                content_hash = generate_content_hash(manifest['content'])
+                print("{} file digest is {}".format(manifest['filename'], content_hash))
```
miteshvp: either we can use a logger error or we can `raise FatalTaskError`

bayesian/dependency_finder.py
```diff
@@ -46,6 +50,18 @@ def execute(self, arguments, db, manifests):
...
+
+                s3 = AmazonS3(bucket_name='boosters-manifest')
+                try:
+                    s3.connect()
+                    manifest['content'] = s3.retrieve_blob(content_hash).decode('utf-8')
+                except ClientError as e:
+                    print("Unexpected error while retrieving S3 data: %s" % e)
```
miteshvp: either we can use a logger error or we can `raise FatalTaskError`

samuzzal-choudhury: @miteshvp this is within server. So, raising FatalTaskError does not make sense. Rather any exception here is going to send an HTTP 500 response.

samuzzal-choudhury:
@tisnik please elaborate on the changes are you looking at?

tisnik:
@samuzzal-choudhury there's new function in utils module and four new blocks of code that needed to be checked a bit.

samuzzal-choudhury:
@tisnik sorry that the coverage results are going bad because of that. I can add the logic of the new function to the API code directly. But, the rest of the changes are necessary ones.

tisnik:
@samuzzal-choudhury NP. actually IIRC all what's needed is one unit tests for the new function. Or -if you merge this PR, I can write the test later if you prefer.

miteshvp:
@tisnik going to merge this to unblock others. Will be really  appreciated if you can add those tests as you suggested. TYVM in advance.
cc @samuzzal-choudhury

PR #308, opened 2018-04-24
Update dependencies

msrb:
Replace unmaintained pycrypto with cryptography.


msrb:
Tested on the devcluster. Seems to be working fine.

PR #325, opened 2018-05-24
Update Dockerfile

humaton:
Use git ls-remote to get commit hash
.git directory is not available during Docker builds in Openshift


humaton:
@samuzzal-choudhury can you have a look at this?
I have changed it because in openshift builds using docker or git strategy we do not have access to .git/ directory. The only thing that I am not sure about is the date. Do we need the date to be part of the service version string?

PR #331, opened 2018-06-06
remove index number from dict manifest_response

ravsa:


miteshvp:
@ravsa - can you please rebase

PR #332, opened 2018-06-07
Endpoint to return core dependencies for each runtime blank booster

sunilk747:


tisnik:
just a question @ravsa . it looks like the following line:

```
fetch_file_from_github_release(CORE_DEPENDENCIES_REPO_URL, 'core.json')
```
means that the json file will be downloaded on every GET requests. Is not it better to have some sort of caching there?

ravsa:
yes  @tisnik, we are planning to add similar caching functionality https://github.com/fabric8-analytics/fabric8-analytics-server/blob/master/bayesian/license_extractor.py#L15

tisnik:
@ravsa nice then. TYVM!

bayesian/utils.py
```diff
@@ -925,3 +927,12 @@ def generate_content_hash(content):
     """Return the sha1 digest of a string."""
     hash_object = hashlib.sha1(content.encode('utf-8'))
     return hash_object.hexdigest()
+
+
+@lru_cache_function(max_size=2048, expiration=60 * 60 * 24)
+def get_core_dependencies(runtime):
+    """Return core dependencies for each runtime."""
+    fetched_file = fetch_file_from_github(CORE_DEPENDENCIES_REPO_URL, 'core.json')
+    dependencies = fetched_file[0].get('content', {})
```
ravsa: `fetched_file_from_github` returns the data in the raw format you need to parse it in JSON format.

tests/test_utils.py
```diff
@@ -216,3 +217,21 @@ def test_is_valid():
     """Check the function is_valid()."""
     assert not is_valid(None)
     assert is_valid("parameter")
+
+
+def test_get_core_dependencies():
+    """Check the function get_core_dependencies()."""
+    result = [
```
ravsa: @sunilk747  Do Not hard-code the values assert for the data exist in the response or not

tests/test_utils.py
```diff
@@ -216,3 +217,21 @@ def test_is_valid():
...
+         "artifactId": "vertx-core",
+         "version": "3.5.0"
+      },
+      {
+         "groupId": "io.vertx",
+         "artifactId": "vertx-unit",
+         "version": "3.5.0"
+      }
+    ]
+    assert get_core_dependencies('vertx') == result
```
ravsa:
```python
assert any(get_core_dependencies('vertx'))
```

bayesian/api_v1.py
```diff
@@ -1076,6 +1076,36 @@ def get(runtime):
...
+class CoreDependencies(ResourceWithSchema):
+    """Implementation of Blank booster /get-core-dependencies REST API call."""
+
+    method_decorators = [login_required]
+
+    @staticmethod
+    def get(runtime):
+        """Handle the GET REST API call."""
+        try:
+            resolved = list()
```
tisnik: `resolved = []` is more idiomatic and faster as well https://stackoverflow.com/questions/5790860/and-vs-list-and-dict-which-is-better

bayesian/api_v1.py
```diff
@@ -1076,6 +1076,36 @@ def get(runtime):
...
+            dependencies = get_core_dependencies(runtime)
+            request_id = uuid.uuid4().hex
+            for elem in dependencies:
+                packages = dict()
```
tisnik: dtto here: `packages = {}`

bayesian/utils.py
```diff
@@ -925,3 +927,12 @@ def generate_content_hash(content):
     """Return the sha1 digest of a string."""
     hash_object = hashlib.sha1(content.encode('utf-8'))
     return hash_object.hexdigest()
+
+
+@lru_cache_function(max_size=2048, expiration=60 * 60 * 24)
```
tisnik: I like this solution!

PR #337, opened 2018-06-22
Fix problems found by shellcheck

tisnik:
Problem description: https://github.com/openshiftio/openshift.io/issues/3751


tisnik:
[test]

PR #341, opened 2018-06-25
added call for scan

yzainee:


bayesian/api_v1.py
```diff
@@ -737,9 +738,23 @@ def post():
         github_url = request.form.get("github_url")
         ref = request.form.get('github_ref')
         user_email = request.headers.get('UserEmail')
+        scan_repo_url = request.headers.get('scanRepoUrl')
+        print('Email ' + user_email)
```
miteshvp: Use logger statements as in - https://github.com/fabric8-analytics/fabric8-analytics-server/blob/master/bayesian/api_v1.py#L761

miteshvp: I would not even want to log this information in our logs. especially user_email

bayesian/api_v1.py
```diff
@@ -737,9 +738,23 @@ def post():
...
+        print('GitUrl ' + scan_repo_url)
         if not user_email:
             user_email = decoded.get('email', 'bayesian@redhat.com')
 
+        if scan_repo_url is not None and scan_repo_url is not "":
```
miteshvp: or use `if scan_repo_url`. It will work exactly as this if.

bayesian/api_v1.py
```diff
@@ -737,9 +738,23 @@ def post():
...
+            try:
+                api_url = GEMINI_SERVER_URL
+                data = {'git-url': scan_repo_url,
+                        'email-ids': [user_email]}
```
miteshvp: do we want to send a list here. I guess there will always be a single email associated to the build?

bayesian/default_config.py
```diff
@@ -49,3 +49,7 @@
 CORE_DEPENDENCIES_REPO_URL = os.getenv('CORE_DEPENDENCIES_REPO', 'https://github.com/'
                                                                  'fabric8-analytics/'
                                                                  'booster-core-dependencies')
+
+GEMINI_SERVER_URL = "http://{host}:{port}".format(
```
miteshvp: You can make use of gremlin url from - https://github.com/fabric8-analytics/fabric8-analytics-server/blob/master/bayesian/utils.py#L41

bayesian/api_v1.py
```diff
@@ -45,6 +45,7 @@
 import os
 from f8a_worker.storages import AmazonS3
 from .generate_manifest import PomXMLTemplate
+from .default_config import GEMINI_SERVER_URL
```
miteshvp: Use gremlin url available in utils

bayesian/api_v1.py
yzainee: @samuzzal-choudhury wants its in a list format

bayesian/default_config.py
miteshvp: I meant, you can also use GEMINI in utils.

PR #354, opened 2018-07-17
Scripts to detect dead code and other issues

tisnik:


tisnik:
Tracking issue https://github.com/openshiftio/openshift.io/issues/3808

PR #358, opened 2018-07-26
Pass dependency files from stack analysis to gemini

abs51295:
Passes 'direct-dependencies.txt' and 'transitive-dependencies.txt' from build to gemini.


abs51295:
[test]

PR #362, opened 2018-08-20
Add a swagger spec with component-analyses endpoint

samuzzal-choudhury:


swagger.yaml
```diff
@@ -0,0 +1,286 @@
...
+      summary: Get job service readiness
+      responses:
+        '200':
+          description: Service is ready
+  '/component-analyses/{ecosystem}/{component}/{version}':
+    get:
+      tags:
+        - Scan Services
+      operationId: f8a_scanner.api_v1.component-analyses
+      summary: Get details about a software component
```
miteshvp: How about something like `Get security and other details about a software component and recommendations to change to a more secured version`

swagger.yaml
```diff
@@ -0,0 +1,286 @@
...
+          schema:
+            $ref: "#/definitions/ComponentAnalysesResponse"
+          description: Detailed response for the requested application component
+        '401':
+          description: Request unauthorized
+        '404':
+          description: Data not found
+definitions:
+  Version:
+    title: Component Version Details
```
miteshvp: Most of the properties are not being maintained. We need cleanup of the response json.

swagger.yaml
```diff
@@ -0,0 +1,286 @@
...
+      shipped_as_downstream:
+        type: array
+        items:
+          type: boolean
+      dependents_count:
+        type: array
+        items:
+          type: integer
+  Package:
+    title: Component Package Specific Details
```
miteshvp: Lots of properties are redundant or not maintained. We need to remove the same. Maybe as part of cleanup exercise and proper json response for comp-analyses.

samuzzal-choudhury:
Merging this approved PR.

PR #367, opened 2018-08-24
RecommenderFeedback: json endpoint update

humaton:


bayesian/utils.py
```diff
@@ -1052,13 +1052,19 @@ def get_recommendation_feedback_by_ecosystem(ecosystem):
...
             feedback_dict = {
                 "recommendation_type": feedback.recommendation_type,
                 "recommended_package_name": feedback.package_name,
                 "feedback": feedback.feedback_type,
-                "input_package_list": feedback.stack_request.dep_snapshot.result.details
+                "input_package_list": dependencies
             }
-            result.append(feedback_dict)
+
+        result.append(feedback_dict)
```
msrb: This doesn't look right to me. Why this change?

bayesian/utils.py
```diff
@@ -1052,13 +1052,19 @@ def get_recommendation_feedback_by_ecosystem(ecosystem):
             filter(RecommendationFeedback.stack_id == StackAnalysisRequest.id).all()
         result = []
         for feedback in feedback_list:
+            if feedback.stack_request.dep_snapshot.result is None:
+                dependencies = []
+            else:
+                dependencies = feedback.stack_request.dep_snapshot.result.details._resolved
```
msrb: If `dep_snapshot` is a dictionary, then this won't work.

`feedback.stack_request.dep_snapshot.get('result', [{}])[0].get('details', [{}])[0].get('_resolved', [])`

^^ that should do the trick, IMO. But I haven't tested it.

bayesian/utils.py
```diff
@@ -1052,11 +1052,19 @@ def get_recommendation_feedback_by_ecosystem(ecosystem):
             filter(RecommendationFeedback.stack_id == StackAnalysisRequest.id).all()
         result = []
         for feedback in feedback_list:
+            if feedback.stack_request.dep_snapshot.result is None:
```
msrb: I think this should be `if feedback.stack_request.dep_snapshot is None`. I.e. without `.result` part.

msrb: Or, even better, `if not feedback.stack_request.dep_snapshot:`

msrb:
[test]

PR #376, opened 2018-09-10
fix bug unused var and redefinition of var

ravsa:
Fix: (https://github.com/openshiftio/openshift.io/issues/4282)


ravsa:
Changes in tests are not required otherwise build could have failed. I guess current test cases are based on the API endpoint name.
Thanks @animuk

PR #395, opened 2018-10-09
Fetch CVE data from CVE nodes

msrb:


bayesian/utils.py
```diff
@@ -277,10 +314,25 @@ def search_packages_from_graph(tokens):
 
 def get_analyses_from_graph(ecosystem, package, version):
     """Read analysis for given package+version from the graph database."""
-    qstring = "g.V().has('ecosystem','" + ecosystem + "').has('name','" + package + "')" \
-              ".as('package').out('has_version').as('version').dedup()." \
-              "select('package', 'version').by(valueMap());"
-    payload = {'gremlin': qstring}
+    script = """\
+results=[];cves=[];nocve_versions=[];\
```
msrb: I am 100% this could be simplified, but I've tried like 100 different things and none of them worked as I wanted. The idea here is that we fetch all package/version data for given EPV, plus CVEs for this EPV. In addition to that, we also return a list of versions which don't have CVEs, so we can recommend them later.

miteshvp: @msrb - you can use below query
```
g.V().has('ecosystem',ecosystem).has('name',name).as('pkg').out('has_version').as('ver').coalesce(out('has_cve').as('cve').select('pkg','ver','cve').by(valueMap()), select('pkg','ver').by(valueMap()));
```

PR #412, opened 2018-11-06
api endpoint for cves bydate&ecosystem

cnulenka:
**[WIP]**Initial changes for public api creation.


bayesian/utils.py
```diff
@@ -370,6 +370,85 @@ def get_analyses_from_graph(ecosystem, package, version):
...
+    cve_nodes_by_date_ecosystem_script_template = """\
+    g.V().has('vertex_label','CVE')\
+    .has('modified_date', modified_date)\
+    .has('ecosystem',ecosystem).as('cve')\
+    .in('has_cve').as('epv')\
+    .select('cve','epv').by(valueMap())\
+    dedup()\
+    """
+
+    def get_cves_by_date():
```
samuzzal-choudhury: @cnulenka how are you planning to unit test these inner functions?

cnulenka: Should I put all these functions in a class? I shall use object of the class to use all these (inner) functions, instead of a call to the outer function.

samuzzal-choudhury: @cnulenka that'll be a better approach in my view.

bayesian/utils.py
```diff
@@ -370,6 +370,84 @@ def get_analyses_from_graph(ecosystem, package, version):
     return resp
 
 
+class GetCveByDateEcosystem:
```
samuzzal-choudhury: @cnulenka please change the class name based on https://visualgit.readthedocs.io/en/latest/pages/naming_convention.html

tests/test_utils.py
```diff
@@ -25,6 +29,39 @@
...
+                    "modified_date": ["20180911"]
+                },
+                "epv": {
+                    "pname": ["io.vertx:vertx-core"],
+                    "version": ["3.4.1"],
+                    "pecosystem": ["maven"],
+                }
+            },
+            {
+                "cve": {
```
miteshvp: `cve` key will not be present. All the block should be moved (outdented)

bayesian/utils.py
```diff
@@ -370,6 +370,97 @@ def get_analyses_from_graph(ecosystem, package, version):
...
+                    "version": cve.get('epv').get('version', [None])[0],
+                    "status": cve.get('cve').get('status', [None])[0],
+                    "fixed_in": cve.get('cve').get('fixed_in', [None])[0],
+                    "link": "https://nvd.nist.gov/vuln/detail/" +
+                            cve.get('cve').get('cve_id', [''])[0]
+                }
+                cve_list_add.append(cve_dict)
+            else:
+                cve_dict = {
+                    "cve_id": cve.get('cve').get('cve_id', [None])[0],
```
miteshvp: Need to change it to
"cve_id": cve.get('cve_id', [None])[0],

bayesian/utils.py
```diff
@@ -370,6 +370,97 @@ def get_analyses_from_graph(ecosystem, package, version):
...
+                    "ecosystem": cve.get('cve').get('ecosystem', [None])[0]
```
miteshvp: Need to change it to 
"ecosystem": cve.get('ecosystem', [None])[0]

miteshvp:
[test]

PR #420, opened 2018-11-15
useless attribute removed for ecosystem

yzainee:


tisnik:
I'm ok with the patch, but for the future changes it would be great to have everything backed up by unit tests so we'll be able to catch issues earlier ;) [I know that I'm repeating myself, sorry]

PR #421, opened 2018-11-17
fix probable e2e tests failure

miteshvp:


miteshvp:
Let's test this on stage

PR #423, opened 2018-11-19
improve npmlist parsing to make it more robust

miteshvp:


miteshvp:
Lets test this on stage

PR #429, opened 2018-12-04
remove schemas endpoint

msehnout:


msehnout:
@tisnik: I've got an ACK from @samuzzal-choudhury via Mattermost chat.

PR #436, opened 2018-12-06
Delete more schema-related stuff

msrb:


msrb:
And we are green, gentlemen :)

PR #443, opened 2018-12-21
support for old pipeline

yzainee:


yzainee:
self merge.. added a check for ecosystem

PR #444, opened 2018-12-26
changing ecosystem to maven. Getting java from build

yzainee:


yzainee:
self merge. renaming ecosystem

PR #454, opened 2019-02-18
fix(BUG): Return `In Progress` if recommender or stackAggregator is None

ravsa:


yzainee:
LGTM. @miteshvp have a look

PR #457, opened 2019-03-06
adding maven support for vscode

yzainee:


Dockerfile
```diff
@@ -2,7 +2,8 @@ FROM registry.centos.org/centos/centos:7
 
 ENV LANG=en_US.UTF-8 \
     F8A_WORKER_VERSION=d403113 \
-    F8A_AUTH_VERSION=5211e23
+    F8A_AUTH_VERSION=5211e23 \
+    F8A_UTILS=33df0cc
```
miteshvp: Is this intentional? See different versions of utils for RHEL and non-RHEL

yzainee: no. it was a typo. rectified.

yzainee: i didnt want to use the latest version and hence changed in dockerfile. forgot to change in rhel file

miteshvp:
LGTM. @yzainee - feel free to merge

PR #470, opened 2019-03-19
changed the component analysis query

yzainee:


yzainee:
@miteshvp The time taken is almost half of what it used to take earlier.

PR #477, opened 2019-04-09
Modified query for component analysis

yzainee:


bayesian/utils.py
```diff
@@ -327,18 +327,19 @@ def search_packages_from_graph(tokens):
...
-.by(valueMap()),select('package','version').by(valueMap())).fill(results);
-[epv:results,recommended_versions:no_cve_versions];\
+    script1 = """\
+g.V().has('pecosystem', ecosystem).has('pname', name).has('version', version).as('version')\
+.in('has_version').dedup().as('package').select('version').coalesce(out('has_cve')\
+.as('cve').select('package','version','cve').by(valueMap()),select('package','version')\
+.by(valueMap()));\
 """
 
+    script2 = """\
```
miteshvp: Get the payload and script in the if loop

bayesian/utils.py
```diff
@@ -347,7 +348,44 @@ def get_analyses_from_graph(ecosystem, package, version):
...
+
+            if not (resp['result'].get('data') and len(resp['result'].get('data')) > 0):
+                # trigger unknown component flow in API for missing package
+                return None
+
+            clubbed_data.append({
+                "epv": resp['result']['data']
+            })
+            second_query_flag = False
+            for first_data in resp['result'].get('data'):
```
miteshvp: you can directly use if 'cve' in resp['result'].get('data')[0]

tisnik:
I'm sorry, but the new code is not covered by unit test.

miteshvp:
@tisnik - good catch. 
@yzainee - Please add the relevant unit test case here

PR #481, opened 2019-04-23
LSP - Recommend highest version if a package has CVE

miteshvp:


miteshvp:
Fixes https://github.com/fabric8-analytics/fabric8-analytics-vscode-extension/issues/292

PR #515, opened 2019-05-15
ingestion API changes

yzainee:


bayesian/utils.py
```diff
@@ -64,17 +65,34 @@ def server_run_flow(flow_name, flow_args):
     :param flow_args: arguments for the flow
     :return: dispatcher ID handling flow
     """
-    current_app.logger.debug('Running flow {}'.format(flow_name))
```
samuzzal-choudhury: @yzainee are you facing any issues in `current_app.logger`?

yzainee: test cases do not work with current_app

bayesian/utils.py
```diff
@@ -49,6 +47,9 @@
     "Do you want to consider adding this Package?"
 
 zero_version = sv.Version("0.0.0")
+AUTH_KEY = os.getenv('OS_AUTH_KEY', '')
```
samuzzal-choudhury: @yzainee is AUTH_KEY used in the code?

yzainee: not currently. Because the ingestion API dont have any security check at present. Once we add that, we would need some sort of authentication to be sent.. or we can remove the route from the ingestion API and make it an internal API call.

arajkumar:
Duplicate of https://github.com/fabric8-analytics/fabric8-analytics-server/pull/705

PR #520, opened 2019-06-13
Enable Prometheus metrics on API Server

miteshvp:


miteshvp:
Let's test this on stage

PR #531, opened 2019-06-19
non cve latest version

yzainee:


bayesian/utils.py
```diff
@@ -349,47 +352,52 @@ def get_analyses_from_graph(ecosystem, package, version):
...
-                        'ecosystem': ecosystem,
-                        'name': package,
-                        'version': version
-                    }
-                }
-
-                graph_req2 = post(gremlin_url, data=json.dumps(payload))
-                if graph_req2 is not None:
-                    resp2 = graph_req2.json()
+            if "cve" in result_data[0]:
```
samuzzal-choudhury: @yzainee is this check enough to conclude that there is a CVE present in the `result.data.cve.cve_id` list?

yzainee: yes, the "cve" will be a part of the json only if the cve exists for that package. The query is written in such a manner

PR #533, opened 2019-06-23
do not install promethues flask exporter twice

miteshvp:


Dockerfile.rhel
```diff
@@ -30,6 +30,7 @@ RUN pushd /coreapi && \
 RUN pip3 install git+https://github.com/fabric8-analytics/fabric8-analytics-worker.git@${F8A_WORKER_VERSION}
 RUN pip3 install git+https://github.com/fabric8-analytics/fabric8-analytics-auth.git@${F8A_AUTH_VERSION}
 RUN pip3 install git+https://github.com/fabric8-analytics/fabric8-analytics-utils.git@${F8A_UTILS}
+RUN pip3 install git+https://github.com/miteshvp/prometheus_flask_exporter.git
```
samuzzal-choudhury: @miteshvp please add your reason here to install `prometheus_flask_exporter` from `miteshvp` and not from `pip`.

miteshvp: It is a custom modification to include moving average response times on top of upstream library.

samuzzal-choudhury: @miteshvp let's plan to add this custom library in fabric8-analytics org and make use of that in the future. For now, it looks good to me.

miteshvp:
Let's test this on stage

PR #534, opened 2019-06-24
remove storage options as ebs storage is behaving erratic

miteshvp:


miteshvp:
Let's test this on stage

PR #545, opened 2019-07-01
time lsp calls and invoke prometheus metrics collector

miteshvp:


bayesian/api_v1.py
```diff
@@ -71,6 +72,11 @@
 TOTAL_COUNT_KEY = 'total_count'
 
 ANALYTICS_API_VERSION = "v1.0"
+HOSTNAME = os.environ.get('HOSTNAME', 'bayesian-api')
+METRICS_SERVICE_URL = "http://{}:{}".format(
+    os.environ.get('METRICS_COLLECTOR_HOST', 'metrics-accumulator'),
```
samuzzal-choudhury: @miteshvp  The environment variables have to be set as `METRICS_ACCUMULATOR_HOST/_PORT` If the service name is `metrics-accumulator`.

openshift/template.yaml
```diff
@@ -36,6 +36,10 @@ objects:
             value: "api"
           - name: F8_API_BACKBONE_HOST
             value: "http://f8a-server-backbone:5000"
+          - name: METRICS_COLLECTOR_HOST
```
samuzzal-choudhury: Openshift will take care of getting these _HOST and _PORT without having an explicit reference in the template as long as that runs in the same project.

openshift/template.yaml
```diff
@@ -36,6 +36,10 @@ objects:
             value: "api"
           - name: F8_API_BACKBONE_HOST
             value: "http://f8a-server-backbone:5000"
+          - name: METRICS_ACCUMULATOR_HOST
+            value: "metrics-accumulator"
+          - name: METRICS_ACCUMULATOR_PORT
+            value: "5200"
```
samuzzal-choudhury: @miteshvp this is redundant. but, it's a nitpick.

PR #560, opened 2019-07-26
API definition changed

yzainee-zz:


swagger.yaml
```diff
@@ -66,21 +66,32 @@ paths:
...
+          required: true
+          type: string
+        - name: origin
+          in: header
+          description: This is required to give an indication as to from where is the request getting generated. If from VScode extension, the value which we need to send is 'vscode'.
+          required: false
+          type: string
+        - name: show_transitive
+          in: header
+          description: This is required to enable or disable the transitive support. We need to pass the value as True if transitive is needed. Default value is False.
```
tisnik: is it case independent please? Also please put 'True'/'False' into apostrophes or quotes, as these values are strings

swagger.yaml
```diff
@@ -66,21 +66,32 @@ paths:
...
+          required: false
```
tisnik: is it required?

tisnik (changes requested):

swagger.yaml
yzainee: Yes this is case sensitive. We need to send 'True'

swagger.yaml
yzainee: By default it takes false as the value. so not required

tisnik (approved):

Issue #550, opened 2019-07-08
Please specify timeout for the Prometheus service HTTP POST call


tisnik:
https://github.com/fabric8-analytics/fabric8-analytics-server/blob/a9330f79207c795366a0f04001b82219d7d73f5f/bayesian/api_v1.py#L272

Most requests to external servers should have a timeout attached, in case the server is not responding in a timely manner. By default, requests do not time out unless a timeout value is set explicitly. Without a timeout, your code may hang for minutes or more.

miteshvp:
It's an async future requests call. API server does not wait for response. Closing this.

tisnik:
@miteshvp is not this session used to call workers asynchronously? I'd suggest to use second FutureSession instance for calling Prometheus

miteshvp:
Actually, its threaded. https://github.com/fabric8-analytics/fabric8-analytics-server/blob/master/bayesian/api_v1.py#L81-L82 so we don't need any other session.

tisnik:
Yup, I meant that we might need to increase the number of workers for the futuresession?

miteshvp:
@tisnik - I guess 100 is a good enough number for our requests. These sessions are not used to invoke workers at the moment. Workers are invoked via another async flow https://github.com/fabric8-analytics/fabric8-analytics-server/blob/master/bayesian/api_v1.py#L278 I guess we are good here.

tisnik:
Yup, and what timeout value to choose? In this case it might be let's say from one to five seconds at most. WDYT?

miteshvp:
@tisnik - this is like a nano-sec call. Timeout doesn't make sense here

tisnik:
@miteshvp yes - in case everything is working as expected. But in reality we'll see network issues that would need to be handled by our service. We can discuss in in a chat rather than there ;)

PR #571, opened 2019-08-07
added check for invalid maven pkgs

yzainee:


yzainee:
[test]

yzainee:
[test]

Dockerfile
```diff
@@ -8,8 +8,8 @@ ENV LANG=en_US.UTF-8 \
 RUN useradd -d /coreapi coreapi
 
 # https://copr.fedorainfracloud.org/coprs/fche/pcp/
-# https://copr.fedorainfracloud.org/coprs/jpopelka/mercator/
-COPY hack/_copr_fche_pcp.repo hack/_copr_jpopelka-mercator.repo /etc/yum.repos.d/
+# https://copr.fedorainfracloud.org/coprs/msrb/mercator/
+COPY hack/_copr_fche_pcp.repo hack/_copr_msrb-mercator.repo /etc/yum.repos.d/
```
msrb: Note mercator from my repo will be newer version than what's in Jirka's repo. Therefore this change could break some code, if we use mercator in server somewhere.

bayesian/api_v1.py
```diff
@@ -263,7 +263,16 @@ def get(ecosystem, package, version):
...
         if ecosystem == 'maven':
-            package = MavenCoordinates.normalize_str(package)
+            try:
+                package = MavenCoordinates.normalize_str(package)
+            except Exception:
+                msg = "Invalid maven format - {pkg}".format(
+                    pkg=package
+                )
+                metrics_payload.update({"status_code": 400, "value": time.time() - st})
+                _session.post(url=METRICS_SERVICE_URL + "/api/v1/prometheus", json=metrics_payload)
```
msrb: This is kinda strange, but if it is implemented this way currently, then let's keep it.

bayesian/api_v1.py
```diff
@@ -263,7 +263,16 @@ def get(ecosystem, package, version):
             )
             raise HTTPError(400, msg)
         if ecosystem == 'maven':
-            package = MavenCoordinates.normalize_str(package)
+            try:
+                package = MavenCoordinates.normalize_str(package)
+            except Exception:
```
msrb: Do we need to catch `Exception` here? I'd hope that MavenCoordinates.normalize_str will throw something like `ValueError`. Catching all exception could hide other errors in future and logs would be full of misleading information about invalid format.

bayesian/api_v1.py
tisnik: Well I agree with @msrb (and made similar comments elsewhere, for example there https://github.com/fabric8-analytics/fabric8-analytics-server/issues/550). The main problem is that the post does not specify timeout, so it will blocks the entire post queue when the Prometheus interface service won't be available for whatever reason.

4 @msrb: IIRC it does not call the real Prometheus, but a new service that are stateful and provides "new" funtionality like computation of moving average etc.

bayesian/api_v1.py
yzainee: changed

PR #573, opened 2019-08-16
api documentation moved from gdoc to github

sivaavkd:


API.md
```diff
@@ -0,0 +1,196 @@
...
+
+- **Sample Responses**
+
+  **_Success_**
+
+    Sample success reponse is [here](https://gist.github.com/sivaavkd/1b27d01ccd17af839c0ba634ce9de628)
+
+  **_In Progress_**
+    ```
+        200:
```
miteshvp: This will be 202. In case of 200, user gets to see the report

miteshvp (changes requested):
Just one minor change. You can add the response for 200 for stack-analyses

API.md
sivaavkd: Thanks Mitesh.  Updated it to 202.

sivaavkd: As you suggested, also added 202 for component analysis.

miteshvp (approved):

API.md
```diff
@@ -0,0 +1,203 @@
+# CodeReady Analytics Core API Specifications
+
+CodeReady Analytics is an open-source analytics platform aimed primarily at assisting developers in getting insights and recommendations on the selected dependency/package stack used for developing their applications.
+
```
Preeticp: replace / with or

API.md
```diff
@@ -0,0 +1,203 @@
...
+This documentation is for developers who are interested in integrating their services or applications with CodeReady Analytics platform.
+
+## **Supported ecosystems**
+
+Below are the ecosystems that CodeReady Analytics currently supports.  We will be adding support for more ecosystems in the near future.
+| | |
```
Preeticp: Currently, CodeReady Analytics supports the following ecosystems. We will add support for more ecosystems in the near future.

API.md
```diff
@@ -0,0 +1,203 @@
...
+|-|-|
+|**Ecosystem**|**Manifest file**|
+| Java Maven | pom.xml |
+| Python Pypi | requirements.txt |
+| NodeJS | package.json |
+
+## **Prerequisites**
+
+Any new integration with the CodeReady Analytics platform will have to do that using an existing 3-scale API management gateway setup. Please reach out to us by raising an issue [here](https://github.com/fabric8-analytics/fabric8-analytics-vscode-extension/issues)
+
```
Preeticp: To integrate with the CodeReady Analytics platform, you need to use an existing 3-scale API management gateway. Create an issue[here](https://github.com/fabric8-analytics/fabric8-analytics-vscode-extension/issues) to reach out to us for the API management gateway.

API.md
```diff
@@ -0,0 +1,203 @@
...
+## **Initiate Dependency Analysis**
+
+Submits a dependency stack analysis request
```
Preeticp: Can we use the format used in this doc here: https://docs.openshift.io/ct-analytics-integration-api-guide.html#ct_api_post_stack_analyses_request

POST /stack-analyses

Purpose: 

Here is the asciidoc source file for it: https://raw.githubusercontent.com/fabric8io/fabric8-online-docs/master/docs/topics/modules/api_post_stack_analyses_request.adoc
Here it is unclear what 'Submits a dependency...'

API.md
```diff
@@ -0,0 +1,203 @@
...
+  `POST`
+
+- **Response Format**
+  `jSON`
+
+- **Path Parameters**
+  `None`
+
+- **Query Parameters**
+  | | | | |
```
Preeticp: There is a problem with rendering tables, see the rendered version: https://github.com/fabric8-analytics/fabric8-analytics-server/blob/0cd38a536bea35a6e3d08e212f3e86ee11b60143/API.md 
Ideallu I would suggest you use Asccidoc format  for the docs but if you want to use mark-down, see here for information on tables in markdown: https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#tables

API.md
```diff
@@ -0,0 +1,203 @@
...
+    ```
+        401:
+            {
+                "error": "Authentication failed - could not decode JWT token"
+            }
+    ```
+
+## **Fetch Dependency Analysis**
+
+Fetches dependency stack analysis response using the request identifier returned by initiate dependency analysis.
```
Preeticp: Fetches the dependency stack analysis response using the request identifier returned by the initiated dependency analysis.

API.md
sivaavkd: We wanted to avoid the usage of stack report initially and rather use the word dependency to make it more understandable for external users, hence the change.

API.md
sivaavkd: Updated purpose as per your suggestion.

API.md
sivaavkd: updated purpose as per your suggestion

Preeticp:
LGTM

PR #577, opened 2019-08-27
Generate empty recommendation for stacks without analyzed dependencies

samuzzal-choudhury:
Related Issue# https://github.com/fabric8-analytics/fabric8-analytics-vscode-extension/issues/253


samuzzal-choudhury:
### After the fix:
image

PR #590, opened 2019-10-30
ecosystem header missing error handled

deepak1725:
Jira: https://jira.coreos.com/browse/APPAI-1035

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ RUN yum install -y epel-release &&\
 
 COPY ./requirements.txt /coreapi/
 RUN pushd /coreapi && \
-    pip3 install -r requirements.txt && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -18,7 +18,7 @@ RUN yum install -y epel-release &&\
 
 COPY ./requirements.txt /coreapi/
 RUN pushd /coreapi && \
-    pip3 install -r requirements.txt && \
+    python3.6 -m pip install -r requirements.txt && \
```
samuzzal-choudhury: @deepak1725 any particular reason for this change?

deepak1725: @samuzzal-choudhury Build is failing. I mean Unit tests are failing without it. 
Please notice that with previous commit.

PR #592, opened 2019-11-29
Updating response code for 202 status

deepak1725:
202 status is not taken as error and hence error is not raised.


bayesian/api_v1.py
```diff
@@ -400,8 +400,8 @@ def get(external_request_id):
                 raise HTTPError(408, "Stack analysis request {t} has timed out. Please retry "
                                      "with a new analysis.".format(t=external_request_id))
             else:
-                raise HTTPError(202, "Analysis for request ID '{t}' is in progress".format(
-                    t=external_request_id))
+                return jsonify("Analysis for request ID '{t}' is in progress".format(
```
samuzzal-choudhury: @deepak1725 Are you adding any unit/integration test steps for asserting these scenarios?

deepak1725: Yes added

PR #593, opened 2019-12-04
change template based on new secret naming convention

miteshvp:


miteshvp:
> @miteshvp what about S3 access key and secret?

@yzainee - I did not see S3 used anywhere in server code. It was used earlier for feedback, but it is moved to RDS now. Hence thought of removing and check in staging once.

yzainee:
ok

PR #600, opened 2020-02-25
query to fetch synk cves in component analyses

deepak1725:
Adding a query to bring synk cves in component analyses from GraphDB

Make sure to add Header: 
`security-vendor = synk`

### DIFF
--- a/bayesian/api_v1.py
+++ b/bayesian/api_v1.py
@@ -31,7 +31,7 @@ from fabric8a_auth.auth import login_required
 from .auth import get_access_token
 from .exceptions import HTTPError
 from .utils import (get_system_version, retrieve_worker_result,
-                    server_create_component_bookkeeping,
+                    server_create_component_bookkeeping, GraphAnalyses,
                     server_create_analysis, get_analyses_from_graph,
                     search_packages_from_graph, fetch_file_from_github_release,
                     get_item_from_list_by_key_value, RecommendationReason,
@@ -275,7 +275,9 @@ class ComponentAnalyses(Resource):
                 raise HTTPError(400, msg)
 
         package = case_sensitivity_transform(ecosystem, package)
-        result = get_analyses_from_graph(ecosystem, package, version, vendor=security_vendor)
+        # Querying GraphDB for CVE Info.
+        result = GraphAnalyses(ecosystem, package, version,
+                               vendor=security_vendor).main()
 
         if result is not None:
             # Known component for Bayesian
--- a/bayesian/api_v1.py
+++ b/bayesian/api_v1.py
@@ -276,8 +276,16 @@ class ComponentAnalyses(Resource):
 
         package = case_sensitivity_transform(ecosystem, package)
         # Querying GraphDB for CVE Info.
-        result = GraphAnalyses(ecosystem, package, version,
-                               vendor=security_vendor).main()
+
+        if security_vendor:
+            graph_obj = GraphAnalyses(ecosystem, package, version, vendor=security_vendor)
+            func = {
+                'snyk': graph_obj.get_analyses_for_snyk
+            }
+            result = eval(func.get(security_vendor))
+        else:
+            # Old Flow
+            result = get_analyses_from_graph(ecosystem, package, version)
 
         if result is not None:
             # Known component for Bayesian
--- a/bayesian/api_v1.py
+++ b/bayesian/api_v1.py
@@ -246,6 +246,7 @@ class ComponentAnalyses(Resource):
     @staticmethod
     def get(ecosystem, package, version):
         """Handle the GET REST API call."""
+        secy_vendor = request.headers.get('secy_vendor', None)
         st = time.time()
         metrics_payload = {
             "pid": os.getpid(),
@@ -274,7 +275,7 @@ class ComponentAnalyses(Resource):
                 raise HTTPError(400, msg)
 
         package = case_sensitivity_transform(ecosystem, package)
-        result = get_analyses_from_graph(ecosystem, package, version)
+        result = get_analyses_from_graph(ecosystem, package, version, vendor=secy_vendor)
 
         if result is not None:
             # Known component for Bayesian
--- a/bayesian/api_v1.py
+++ b/bayesian/api_v1.py
@@ -246,7 +246,7 @@ class ComponentAnalyses(Resource):
     @staticmethod
     def get(ecosystem, package, version):
         """Handle the GET REST API call."""
-        security_vendor = request.headers.get('security_vendor', None)
+        security_vendor = request.headers.get('security-vendor', None)
         st = time.time()
         metrics_payload = {
             "pid": os.getpid(),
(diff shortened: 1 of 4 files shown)


bayesian/api_v1.py
```diff
@@ -246,6 +246,7 @@ class ComponentAnalyses(Resource):
     @staticmethod
     def get(ecosystem, package, version):
         """Handle the GET REST API call."""
+        secy_vendor = request.headers.get('secy_vendor', None)
```
yzainee: Either use "security_vendor" or use "sec_vendor"

bayesian/utils.py
```diff
@@ -329,14 +329,23 @@ def search_packages_from_graph(tokens):
...
-def get_analyses_from_graph(ecosystem, package, version):
+def get_analyses_from_graph(ecosystem, package, version, vendor=None):
     """Read analysis for given package+version from the graph database."""
     script1 = """\
-g.V().has('pecosystem', ecosystem).has('pname', name).has('version', version).as('version')\
-.in('has_version').dedup().as('package').select('version').coalesce(out('has_cve')\
-.as('cve').select('package','version','cve').by(valueMap()),select('package','version')\
-.by(valueMap()));\
-"""
+        g.V().has('pecosystem', ecosystem).has('pname', name).has('version', version).as('version')\
```
yzainee: Instead of using if else, use a separate function so that later if we need to add any filter etc, we can do it in 1 function

bayesian/utils.py
```diff
@@ -369,9 +378,7 @@ def get_analyses_from_graph(ecosystem, package, version):
                         "recommended_versions": result_data[0]['package']['latest_non_cve_version']
                     })
                 else:
-                    script2 = "g.V().has('pecosystem', ecosystem).has('pname', name)" \
-                              ".has('version', version).in('has_version')" \
-                              ".out('has_version').not(out('has_cve')).values('version').dedup();"
```
yzainee: It was added here as a part of the review comment. Instead of doing it every time, we do it only for else case.

bayesian/utils.py
```diff
@@ -329,14 +329,23 @@ def search_packages_from_graph(tokens):
...
+        .as('cve').select('package','version','cve').by(valueMap()),select('package','version')\
+        .by(valueMap()));\
+        """
+
+    script2 = \
+        "g.V().has('pecosystem', ecosystem).has('pname', name)" \
+        ".has('version', version).in('has_version')" \
+        ".out('has_version').not(out('has_cve')).values('version').dedup();"
+
+    if vendor:
```
yzainee: what happens when no vendor is provided? for the current ext users? This logic needs to be changed

deepak1725: if vendor is None.. Flow Continues..skipping Condition. Please relook

yzainee:
By the way, you cannot merge this PR anyways.. not until the graph schema etc is all in place

bayesian/utils.py
deepak1725: It is Okay to declare the variable at function start. another option is to put a conditional statement here... which increases cyclometric complexity and hence it holds No value gains at all. 
Declaring `script2` at the start...has No performance loss at all.

deepak1725:
> By the way, you cannot merge this PR anyways.. not until the graph schema etc is all in place

Merging will have No effect, as in order to execute conditional flow.. all we need is `secy_vendor` in headers. 
Passing of E2E Tests reflects we will have no problem at all.

bayesian/api_v1.py
deepak1725: `secy` is an acronym for security. Either way, I am okay to change it to `security`

yzainee:
In case of snyk, we dont need to fetch the data of version from the pkg node. We will get that in the cve data itself. Thats why im saying, create 2 separate functions, else you will end up with if else logic everywhere

bayesian/utils.py
yzainee: i got it.  what im trying to say is, you need a separate function call where it does all the fetching etc for snyk cves.

bayesian/utils.py
yzainee: We dont need any conditions here. script2 variable gets initialized only when there is no value present in the recommended versions. as we are dealing with 0.8 seconds of response time, we dont want to risk any additional logic which is not used later. same is the case with multiple if checks etc. Thats why, im saying create 2 separate function flows. call function x in normal case, function y for snyk case

bayesian/utils.py
deepak1725: Added

bayesian/utils.py
deepak1725: added separate function

bayesian/utils.py
deepak1725: Added Separate Query Builder Function

deepak1725:
> In case of snyk, we dont need to fetch the data of version from the pkg node. We will get that in the cve data itself. Thats why im saying, create 2 separate functions, else you will end up with if else logic everywhere. The 2nd query itself is not needed for snyk

Done

bayesian/utils.py
```diff
@@ -329,8 +329,118 @@ def search_packages_from_graph(tokens):
...
+    def build_graphdb_query(vendor):
+        """Func returns GraphDB query based on vendor argument."""
+        get_cve_info = """
+            g.V().has('pecosystem', ecosystem).has('pname', name).has('version', version)
+            .as('version').in('has_version').dedup().as('package').select('version')
+            .coalesce(out('has_cve').as('cve').select('package','version','cve')
+            .by(valueMap()),select('package','version').by(valueMap()));
+            """
+
+        get_latest_non_cve_version = """
```
yzainee: We dont need this query in the vendor = snyk case. Only the prev query is sufficient

bayesian/utils.py
```diff
@@ -329,8 +329,118 @@ def search_packages_from_graph(tokens):
...
+
+    def main(self):
+        """Fetch analysis for given package+version from the graph database.
+
+        It finally builds and returns JSON Response. This is main Function.
+        """
+        if self.vendor is None:
+            # if vendor is None, resume Old Flow for fetching data from our GraphDB Node
+            return get_analyses_from_graph(self.ecosystem, self.package, self.version)
+        cve_info_query, latest_non_cve_version_query = self.build_graphdb_query(self.vendor)
```
yzainee: The whole logic from here on can go into another function like say 'get_analyses_for_snyk"

bayesian/utils.py
```diff
@@ -329,8 +329,118 @@ def search_packages_from_graph(tokens):
...
+            clubbed_data.append({
+                "epv": result_data
+            })
+
+            if "cve" not in result_data[0]:
+                clubbed_data.append({
+                    "recommended_versions": []
+                })
+
+            if "latest_non_cve_version" in result_data[0]['package']:
```
yzainee: couple of points here
1. this all should go in the else case, i.e when cve information is present in the data
2. you wont get any field like latest_no_cve_version in this case. Change the logic here

bayesian/utils.py
```diff
@@ -329,8 +329,118 @@ def search_packages_from_graph(tokens):
...
+                clubbed_data.append({
+                    "recommended_versions": result_data[0]['package']['latest_non_cve_version']
+                })
+            else:
```
yzainee: there is no such query for latest non cve version in case of snyk

deepak1725:
Unit Tests are added and E2E Tests can be written only after the above changes are deployed on Stage.
Hence we are good to Merge this after Graph Schema updated on Stage.

yzainee:
unit test dont need any graph schema changes. As for testing, it can be done on the dev cluster as well. Please test on dev cluster with the image and queries provided to populate the data

PR #602, opened 2020-03-18
changing data field in sync with graph fields

deepak1725:
Snyk Integration: Changing the Data fields in sync with Graph Changes

### DIFF
--- a/bayesian/utils.py
+++ b/bayesian/utils.py
@@ -381,12 +381,20 @@ class GraphAnalyses:
 
             if "cve" in result_data[0]:
                 # if cve if present
+                rec_version = result_data[0].get('package', {}).get(
+                                                'latest_non_cve_version', {})
                 clubbed_data.append({
-                    "recommended_versions": result_data[0].get('package', {}).get(
-                                                        'latest_non_cve_version', {}),
+                    "recommended_versions": rec_version,
                     "snyk_pkg_link": self.get_link(),
                 })
+                logger.info("latest non cve version for {eco} {pkg} is {ver}".format(
+                    eco=self.ecosystem,
+                    pkg=self.package,
+                    ver=rec_version
+                ))
             else:
+                logger.info("Unable to find recommended version for Package Object:\n{}\n".format(
+                                result_data[0].get('package', {})))
                 clubbed_data.append({
                     "recommended_versions": []
                 })
--- a/bayesian/utils.py
+++ b/bayesian/utils.py
@@ -393,8 +393,6 @@ class GraphAnalyses:
                     ver=rec_version
                 ))
             else:
-                logger.info("Unable to find recommended version for Package Object:\n{}\n".format(
-                                result_data[0].get('package', {})))
                 clubbed_data.append({
                     "recommended_versions": []
                 })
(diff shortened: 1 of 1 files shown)


bayesian/utils.py
```diff
@@ -382,8 +382,9 @@ def get_analyses_for_snyk(self):
             if "cve" in result_data[0]:
                 # if cve if present
                 clubbed_data.append({
-                    "recommended_versions": result_data[0]['cve']['sfixed_in'],
-                    "link": self.get_link(),
+                    "recommended_versions": result_data[0].get('package', {}).get(
```
yzainee: have a if condition here to get the data from latest_non_cve_version, if not found, send blank and add a logger.error msg. Dont add exception. just an error msg

yzainee: something like this
rec_version = ""
if "latest_non_cve_version" in result_data[0]['package']:
 rec_version = bla bla
else
log msg
clubbed_data.appen ........

deepak1725: @yzainee Logic covers missing `latest_non_cve_version` field.
I mean 
```
"recommended_versions": result_data[0].get('package', {}).get(
                                                        'latest_non_cve_version', {}),
```
will return `{}` if there is missing `latest_non_cve_version`

yzainee: i know the logic covers that. i only want a msg in logs for such pkgs so that investigation can be done if needed. right now we wont know why it returned blank

yzainee: whichever way you want, just add a msg if you dont find this value.. in msg just have eco pkg. should be enough

yzainee: or if you dont want to add an if condition, you can add a generic logger.info("latest non cve version for {eco} {pkg} is {ver}") . then whether +ve or -ve, we will see this msg in logs

deepak1725: Sure. I am actually reluctant in adding any conditional statement here... as it will Increase Cyclomatic complexity,which hence make unit Testing a bit harder and may cause Coverage to reduce.

Sure. Going ahead with adding Logs

bayesian/utils.py
```diff
@@ -381,11 +381,20 @@ def get_analyses_for_snyk(self):
...
+                    "recommended_versions": rec_version,
+                    "snyk_pkg_link": self.get_link(),
                 })
+                logger.info("latest non cve version for {eco} {pkg} is {ver}".format(
+                    eco=self.ecosystem,
+                    pkg=self.package,
+                    ver=rec_version
+                ))
             else:
+                logger.info("Unable to find recommended version for Package Object:\n{}\n".format(
```
arajkumar: should be logger.error/warning?

yzainee: @deepak1725 This else logger is printing incorrect msg. The else part is for CVE. so basically whenever there is no cve, it will print this msg. remove it.

deepak1725: Ack!

deepak1725: Removed

bayesian/utils.py
```diff
@@ -381,10 +381,17 @@ def get_analyses_for_snyk(self):
 
             if "cve" in result_data[0]:
                 # if cve if present
+                rec_version = result_data[0].get('package', {}).get(
```
arajkumar: Can you rename variable according to db field name? It should be `latest_non_cve_version` in this case.

deepak1725: its recommended version.

PR #604, opened 2020-03-19
Component analyses changes

deepak1725:
Snyk Integration: Vendor-Specific API Response Changes.

### DIFF
--- a/bayesian/utils.py
+++ b/bayesian/utils.py
@@ -50,7 +50,7 @@ gremlin_url = "http://{host}:{port}".format(
     port=os.environ.get("BAYESIAN_GREMLIN_HTTP_SERVICE_PORT", "8182"))
 
 companion_reason_statement = "along with the provided input stack. " \
-    "Do you want to consider adding this Package?"
+                             "Do you want to consider adding this Package?"
 
 zero_version = sv.Version("0.0.0")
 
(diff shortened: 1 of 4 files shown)


bayesian/utils.py
```diff
@@ -420,6 +472,44 @@ def get_link(self):
...
+            gremlin_resp = SnykComponentAnalysisResponse(package, version, clubbed_data)
+
+            reco['data'] = gremlin_resp.results
+
+            if not gremlin_resp.has_cves():
+                # If there is No CVE, we have No Recommendations for you.
+                reco['recommendation'] = {}
+                return {"result": reco}
+
+            recommendation_dict = reco['recommendation']
```
yzainee: I would suggest you use .get instead of directly accessing values. If some field is missing, you will get an exception

bayesian/utils.py
```diff
@@ -420,6 +472,44 @@ def get_link(self):
...
+            recommendation_dict['component-analyses']['cve'] = gremlin_resp.get_cve_maps()
```
yzainee: will there be a case when component-analyses is not present? If yes, this step might fail

yzainee:
@deepak1725 This review is done keeping the current requirement in mind. This is going to change very soon. Not sure if we want to ship this to stage

bayesian/utils.py
deepak1725: That will never be a case. 
`component-analyses` is picked up from Static Template

bayesian/utils.py
deepak1725: It will never be a case. 
`reco['recommendation']` is picked up from static template

deepak1725:
> @deepak1725 This review is done keeping the current requirement in mind. This is going to change very soon. Not sure if we want to ship this to stage

@yzainee Its completely safe to promote this to stage.
This PR is in line with 
https://github.com/fabric8-analytics/fabric8-analytics-server/pull/600 and 
https://github.com/fabric8-analytics/fabric8-analytics-server/pull/602

Please note this PR only add vendor-specific functionality. Not effecting existing functionality at any level.

deepak1725:
Closing as we have a Retraced path.

PR #605, opened 2020-03-24
New Integration swagger Update

deepak1725:
Updates in API Docs.
* Adding Swagger Update in Sync with New Vendor Integration.
* Making Swagger Interactive.
* Adding Token Authentication. 

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -403,9 +403,9 @@ definitions:
         $ref: "#/definitions/CveDetails"
       message:
         type: string
-  SnykPremiumRecommendation:
-    title: Snyk Premium Recommendation Information
-    description: Snyk Premium Recommendation information
+  SnykRecommendation:
+    title: Snyk Recommendation Information
+    description: Snyk Recommendation information for Standard and Premium Recommendation.
     properties:
       change_to:
         type: string
@@ -416,31 +416,11 @@ definitions:
       severity:
         type: string
       public_vulnerabilities:
-        type: string
+        type: integer
       private_vulnerabilities:
         type: integer
       exploitable_vulnerabilities:
         type: integer
-      snyk_package_link:
-        type: string
-  SnykStandardRecommendation:
-    title: Snyk Recommended Information
-    description: Snyk Recommended information
-    properties:
-      change_to:
-        type: string
-      component-analyses:
-        $ref: "#/definitions/CveDetails"
-      message:
-        type: string
-      severity:
-        type: string
-      public_vulnerabilities:
-        type: string
-      private_vulnerabilities:
-        type: string
-      exploitable_vulnerabilities:
-        type: string
       snyk_package_link:
         type: string
   ComponentAnalysesData:
@@ -454,8 +434,7 @@ definitions:
       recommendation:
         oneOf:
           - $ref: "#/definitions/Recommendation"
-          - $ref: "#/definitions/SnykPremiumRecommendation"
-          - $ref: "#/definitions/SnykStandardRecommendation"
+          - $ref: "#/definitions/SnykRecommendation"
   ComponentAnalysesResponse:
     title: Response of Component Analyses
     description: Component analyses response
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -51,7 +51,7 @@ paths:
           type: string
           required: true
           description: component version
-        - in: query
+        - in: header
           name: security-vendor
           type: string
           required: false
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -14,7 +14,7 @@ securityDefinitions:
     type: apiKey
     name: user_key
     in: query
-    description: Token Auth is used for Authentication
+    description: 3Scale Token Auth is used for Authentication
 
 security:
   - tokenAuth: []
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -1,22 +1,25 @@
 swagger: '2.0'
 info:
   title: Analytics Core API
-  version: '0.1'
+  version: '0.14'
 basePath: /api/v1/
+host: "f8a-analytics-2445582058137.production.gw.apicast.io"
 consumes:
   - application/json
 produces:
   - application/json
+
+securityDefinitions:
+  tokenAuth:
+    type: apiKey
+    name: user_key
+    in: query
+    description: Token Auth is used for Authentication
+
+security:
+  - tokenAuth: []
+
 paths:
-  /liveness:
-    get:
-      tags:
-        - Service settings
-      operationId: f8a_admin.api_v1.get_liveness
-      summary: Get job service liveness
-      responses:
-        '200':
-          description: Service is alive
   /readiness:
     get:
       tags:
@@ -48,6 +51,11 @@ paths:
           type: string
           required: true
           description: component version
+        - in: query
+          name: security-vendor
+          type: string
+          required: false
+          description: Security Vendor
       responses:
         '200':
           schema:
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -1,7 +1,7 @@
 swagger: '2.0'
 info:
   title: Analytics Core API
-  version: '0.14'
+  version: '0.2'
 basePath: /api/v1/
 host: "f8a-analytics-2445582058137.production.gw.apicast.io"
 consumes:
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -403,6 +403,46 @@ definitions:
         $ref: "#/definitions/CveDetails"
       message:
         type: string
+  SnykPremiumRecommendation:
+    title: Snyk Premium Recommendation Information
+    description: Snyk Premium Recommendation information
+    properties:
+      change_to:
+        type: string
+      component-analyses:
+        $ref: "#/definitions/CveDetails"
+      message:
+        type: string
+      severity:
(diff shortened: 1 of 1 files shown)


swagger.yaml
```diff
@@ -48,6 +51,11 @@ paths:
           type: string
           required: true
           description: component version
+        - in: query
```
arajkumar: security-vendor is a header field not query param.

swagger.yaml
```diff
@@ -1,22 +1,25 @@
...
+    type: apiKey
+    name: user_key
+    in: query
+    description: Token Auth is used for Authentication
+
+security:
+  - tokenAuth: []
+
 paths:
-  /liveness:
```
arajkumar: Is there a reason for removing liveness end point from spec. I agree, it is not useful from the documentation perspective.

swagger.yaml
```diff
@@ -1,22 +1,25 @@
...
-  version: '0.1'
+  version: '0.2'
 basePath: /api/v1/
+host: "f8a-analytics-2445582058137.production.gw.apicast.io"
 consumes:
   - application/json
 produces:
   - application/json
+
+securityDefinitions:
```
arajkumar: I guess @yzainee is the right person to add doc related to Auth tokens.

swagger.yaml
deepak1725: @arajkumar Liveness is removed because it is not defined in 3Scale, So, whenever user tries Interactive Mode, he will not be able to get 200 from this endpoint

swagger.yaml
deepak1725: Ack!

swagger.yaml
deepak1725: This Again added to realise Interactive Mode. I have Tested it. Working fine

swagger.yaml
```diff
@@ -1,22 +1,25 @@
...
+  tokenAuth:
+    type: apiKey
+    name: user_key
+    in: query
+    description: Token Auth is used for Authentication
```
arajkumar: It would be good if the `description`  explicitly mentions it deals with 3scale token.

swagger.yaml
```diff
@@ -48,6 +51,11 @@ paths:
           type: string
           required: true
           description: component version
+        - in: header
+          name: security-vendor
+          type: string
+          required: false
```
arajkumar: It would be good to express vendor value as enum,

`enum: [snyk]`, it would help the API user to understand about possible value to this header field.

swagger.yaml
deepak1725: Ack!

swagger.yaml
deepak1725: We are supporting Single Vendor only, No plans in future too of integrating another.
Expecting `string` only in API

arajkumar: What is the use of this field if it is going to be a single vendor and it's value is also fixed. IIRC, I raised this point and @SriKrishnaPaparaju mentioned that it will help be useful to distinguish multiple vendors.

deepak1725: > What is the use of this field if it is going to be a single vendor and its value is also fixed. 
For Differentiating between Vendor and No Vendor, This Helps in selecting the vendor-specific node.
> IIRC, I raised this point and @SriKrishnaPaparaju mentioned that it will help be useful to distinguish multiple vendors.
Multiple vendors still can be integrated, but for an API call, we support a single vendor response at a time.

arajkumar: Thats the reason for asking to add enum. It shows possible accepted values and adds puts constrains on value which you could pass.

deepak1725: We already had this discussion with Krishna and Frontend is already inline with this.

swagger.yaml
```diff
@@ -404,7 +452,10 @@ definitions:
         items:
           $ref: "#/definitions/ComponentAnalysesDataDetails"
       recommendation:
-        $ref: "#/definitions/Recommendation"
+        oneOf:
+          - $ref: "#/definitions/Recommendation"
+          - $ref: "#/definitions/SnykPremiumRecommendation"
+          - $ref: "#/definitions/SnykStandardRecommendation"
```
arajkumar: I don't see any difference between `SnykPremiumRecommendation ` and `SnykStandardRecommendation`  at least from the properties point of view., both same fields.

deepak1725: if observed carefully, you will find difference in `private_vulnerabilities: type: integer` and `exploitable_vulnerabilities: type: integer`

arajkumar: Okay, I see a type difference! 
Can you help me to understand this by giving a sample response for those fields?

deepak1725: Refer Example in Doc

arajkumar: So type `string` for `private_vulnerabilities` and `exploitable_vulnerabilities` is used to represent NA?

deepak1725: yes

arajkumar: If so, instead of using "NA", why can't you use an integer value (say -1) to denote it unavailable. This pattern is followed in the existing response. e.g. `transitive_count`

deepak1725: Okay then, if you are NOT okay with NA, then Two other Possible Options:
1. Completely Omit the field for Standard User.
2. Send -1 for Premium Fields in Standard Response.
Let me know what makes sense.

arajkumar: I prefer `-1` as the pattern is already used in the existing response. @yzainee, Please comment.

deepak1725: Ack!

yzainee:
@deepak1725 referring to these API
'/epvs/bycveid/{cve_id}':
'/cves/bydate/{date}/{ecosystem}':

swagger.yaml
arajkumar: It is going to be still string, but enum will make it clear that all expected value for this field.

swagger.yaml
```diff
@@ -1,22 +1,25 @@
 swagger: '2.0'
 info:
   title: Analytics Core API
-  version: '0.1'
+  version: '0.2'
```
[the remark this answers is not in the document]
deepak1725: Ack! 
Same is practised. Upgraded version for New Updates.

swagger.yaml
```diff
@@ -1,22 +1,25 @@
...
 paths:
-  /liveness:
-    get:
-      tags:
-        - Service settings
-      operationId: f8a_admin.api_v1.get_liveness
-      summary: Get job service liveness
-      responses:
-        '200':
-          description: Service is alive
```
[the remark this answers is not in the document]
deepak1725: These Changes in No matter should have any impact on SLO's.

swagger.yaml
```diff
@@ -395,6 +403,26 @@ definitions:
...
+        $ref: "#/definitions/CveDetails"
+      message:
+        type: string
+      severity:
+        type: string
+      public_vulnerabilities:
+        type: integer
+      private_vulnerabilities:
+        type: integer
+      exploitable_vulnerabilities:
```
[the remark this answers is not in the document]
deepak1725: > what does 'exploitable' mean
We have Vulnerabilities in severity range: 
[Not Defined, Unproven, Proof of Concept, Functional, High]
All Vulnerabilities with severity:
`[functional, high, proof of concept]` will be treated as Exploitable.

In Component Analyses, We are showing a message:
`4 Vulnerabilities with this dependency, 3 with exploitable vulnerabilities.`
 This Param will help in Building this Message.

swagger.yaml
```diff
@@ -404,7 +432,9 @@ definitions:
         items:
           $ref: "#/definitions/ComponentAnalysesDataDetails"
       recommendation:
-        $ref: "#/definitions/Recommendation"
+        oneOf:
+          - $ref: "#/definitions/Recommendation"
+          - $ref: "#/definitions/SnykRecommendation"
```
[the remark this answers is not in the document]
deepak1725: We have now 2 Response Modals
`$ref` is used as a reference to anyOne of them.

swagger.yaml
```diff
@@ -395,6 +403,26 @@ definitions:
...
+        type: string
+      severity:
+        type: string
+      public_vulnerabilities:
+        type: integer
+      private_vulnerabilities:
+        type: integer
+      exploitable_vulnerabilities:
+        type: integer
+      snyk_package_link:
```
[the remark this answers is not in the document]
deepak1725: Response for Registered User and Non-Register User is nearly the same in terms of Response Model. 
Premium Fields like `private_vulnerabilities`, `exploitable_vulnerabilities`, `patch_exists` have same response data types for each Response (Registered / Non Registered). Only difference is they will have value as `-1` instead of having some concreate values.

swagger.yaml
```diff
@@ -48,6 +51,11 @@ paths:
           type: string
           required: true
           description: component version
+        - in: header
+          name: security-vendor
```
yzainee, line 55: If we are having a completely different endpoint (v2) then in that case do we need to pass this value?

yzainee: As we are planning to have the uuid generation at the server side, we should make sure that the component analysis too sends the uuid in its header. Initially it should send blank, and the same would be generated and sent back as the comp analysis response. The uuid can then be saved locally and sent in the subsequent calls.

PR #606, opened 2020-03-24
Stack report API spec changes for snyk

arajkumar:
https://app.swaggerhub.com/apis/arajkumar/stack-analyses_api/2.0


arajkumar:
Swagger 2.0 to OpenAPI 3.0 conversion has been done using https://github.com/Mermade/oas-kit.

OpenAPI 3.0 has some new constructs to express multiple types in response model. Reference: https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/#oneof

deepak1725:
@arajkumar Can you please paste your response here
https://docs.google.com/document/d/1lMlrQBnLVROLlxorBE1cSecpDSN1VUldE9zh921oOAY/edit#
This way it will be helpful to verify response model

swagger.yaml
```diff
@@ -368,15 +368,77 @@ definitions:
...
+          - high
+      title:
+        type: string
+      url:
+        type: string
+      private:
+        type: boolean
+      premium_details:
+        $ref: "#/definitions/SnykCVEPremium"
   CVE:
```
deepak1725: I see definition updates only. 
How are you planning to differentiate b/w API calls from VS Code v0.13 and v.14?

arajkumar: It is for stack report UI which is a hosted Angular App., it is not tied with vscode plugin. Following incremental updates will be done to ensure smooth transition,

- Modify stack report API to return both CVE model and SnykCVE model using Snyk data, this ensures existing stack report UI won't break abruptly.
- Stack report UI will be modified to accommodate snyk specific changes

deepak1725: On similar cases, In Component Analyses, We are passing `security-vendor` header in API calls. 
Can we maintain the same practice here..?

arajkumar: Thanks, I added the same. Please check.

deepak1725:
@arajkumar We discussed Separate Files for Separate Versions..? Right..?

v2/swagger.yaml
```diff
@@ -434,12 +491,26 @@ definitions:
...
+        type: array
+        items:
+          $ref: "#/definitions/Vulnerability"
+        description: Publicly known vulnerability details
+      private_vulnerabilities:
         type: array
+        items:
+          $ref: "#/definitions/Vulnerability"
+        description: Private vulnerability details, available only to registered users
+      affected_direct_deps:
```
arajkumar: It is an existing field which is already used in production, but not present in the document.

deepak1725:
@arajkumar Any reason for not including v2 folder in `bayesian` folder..? I am asking because all our code is in `bayesian` folder. 
Including v2 folder in that might help in shifting codebase for v2 too in future..? 
Just a thought.! WDYT?

v2/swagger.yaml
```diff
@@ -0,0 +1,629 @@
+swagger: '2.0'
+info:
+  title: Analytics Core API
+  version: '0.2'
```
deepak1725, line 4: Can we update version to **2**

v2/swagger.yaml
```diff
@@ -0,0 +1,629 @@
...
+produces:
+  - application/json
+paths:
+  '/component-analyses/{eco}/{name}/{ver}':
+    get:
+      tags:
+        - Scan Services
+      operationId: f8a_scanner.api_v1.component-analyses
+      summary: Get security and other details about a software component and recommendations to change to a more secured version
+      parameters:
```
SriKrishnaPaparaju, line 17: How requests from  'free tier' and 'already registered tier' of users are differentiated ?

v2/swagger.yaml
```diff
@@ -0,0 +1,629 @@
...
+          description: Response with the unique identifier for the stack analysis request
+        '401':
+          description: Request Unauthorized
+        '400':
+          description: Bad request
+  '/stack-analyses/{request-id}':
+    get:
+      tags:
+        - Scan Services
+      operationId: f8a_scanner.api_v1.stack-analyses-get
```
SriKrishnaPaparaju, line 90: Why is this ID referring to v1 ?

v2/swagger.yaml
```diff
@@ -0,0 +1,629 @@
...
+      last_updated:
+        type: array
+        items:
+          type: number
+          format: float
+      vertex_label:
+        type: array
+        items:
+          type: string
+      cve_ids:
```
SriKrishnaPaparaju, line 143: 'vulnerability' might be the right tag to represent this field

v2/swagger.yaml
```diff
@@ -0,0 +1,629 @@
...
+        type: array
+        items:
+          type: string
+      description:
+        type: array
+        items:
+          type: string
+      cm_loc:
```
SriKrishnaPaparaju, line 151: What is the role of this field ?

v2/swagger.yaml
```diff
@@ -0,0 +1,629 @@
...
+      licenses:
+        type: array
+        items:
+          type: string
+      last_updated:
+        type: array
+        items:
+          type: number
+          format: float
+      vertex_label:
```
SriKrishnaPaparaju, line 139: What is the role of this field ?

v2/swagger.yaml
```diff
@@ -0,0 +1,629 @@
...
+          type: string
+      description:
+        type: array
+        items:
+          type: string
+      cm_loc:
+        type: array
+        items:
+          type: integer
+      cm_num_files:
```
SriKrishnaPaparaju, line 155: What is the role of this field ?

v2/swagger.yaml
```diff
@@ -0,0 +1,629 @@
...
+        type: array
+        items:
+          type: integer
+      cm_avg_cyclomatic_complexity:
```
SriKrishnaPaparaju, line 159: Do we show this field as part of UI ?

v2/swagger.yaml
```diff
@@ -0,0 +1,629 @@
...
+        type: array
+        items:
+          type: integer
+      shipped_as_downstream:
```
SriKrishnaPaparaju, line 163: do we display  this field  ?

v2/swagger.yaml
```diff
@@ -0,0 +1,629 @@
...
+        type: array
+        items:
+          type: boolean
+      dependents_count:
```
SriKrishnaPaparaju, line 167: What is the role of this field ?

v2/swagger.yaml
arajkumar: I'm not too much worried about internal version field. I'm fine with 2 also.

arajkumar:
> Any reason for not including v2 folder in bayesian folder..?
We can add source code inside v2/bayesian directory, spec shall be still in v2/.

v2/swagger.yaml
```diff
@@ -0,0 +1,629 @@
...
+          type: string
+  ComponentAnalysesDataDetails:
+    title: Component Analyses Package and Version
+    description: Component analyses data for package and version
+    properties:
+      version:
+        $ref: "#/definitions/Version"
+      package:
+        $ref: "#/definitions/Package"
+  PremiumVulnerabilityFields:
```
SriKrishnaPaparaju, line 292: Does this indicate private vulnerabilities ?

v2/swagger.yaml
```diff
@@ -0,0 +1,629 @@
...
+          - medium
+          - high
+      title:
+        type: string
+      url:
+        type: string
+      premium_details:
+        description: Premium details about a Vulnerability, available only to registered users
+        $ref: "#/definitions/PremiumVulnerabilityFields"
+  CVE:
```
SriKrishnaPaparaju, line 349: If this indicates 'public vulnerability', we need appropriate name for this field

v2/swagger.yaml
SriKrishnaPaparaju: How do we differentiate public and private vulnerabilities ?

v2/swagger.yaml
```diff
@@ -0,0 +1,629 @@
...
+        - Scan Services
+      operationId: f8a_scanner.api_v1.stack-analyses-get
+      summary: Get security and other details about an application stack and related recommendations to improve the stack
+      parameters:
+        - in: path
+          name: request-id
+          type: string
+          required: true
+          description: stack analysis request id
+      responses:
```
SriKrishnaPaparaju, line 98: can we have clear separation of response models to represent free / registered user personas ? These two different response models can in turn reuse any common fields

v2/swagger.yaml
```diff
@@ -0,0 +1,629 @@
...
+        type: array
+        items:
+          $ref: "#/definitions/RecommendedPackageData"
+      manifest_file_path:
+        type: string
+      usage_outliers:
+        type: array
+        items:
+          type: object
+      missing_packages_pgm:
```
SriKrishnaPaparaju, line 590: where do display this field ? can we remove the fields that not are not required ?

v2/swagger.yaml
arajkumar: Authorization token(UUID) will be added once we are done with Authentication flow design.

v2/swagger.yaml
arajkumar: I will fix it in the next patch.

v2/swagger.yaml
arajkumar: I will address in next iteration.

v2/swagger.yaml
arajkumar: Legacy stuff, carried over from v1. I will fix it.

v2/swagger.yaml
arajkumar: Legacy stuff, carried over from v1. I will fix it.

v2/swagger.yaml
arajkumar: Legacy stuff, carried over from v1. I will fix it.

v2/swagger.yaml
arajkumar: Legacy stuff, carried over from v1. I will fix it.

v2/swagger.yaml
arajkumar: Nope, it is not been referred from UI. Legacy stuff, carried over from v1. I will fix it.

v2/swagger.yaml
arajkumar: Nope, it is not been referred from UI. Legacy stuff, carried over from v1. I will fix it.

v2/swagger.yaml
arajkumar: It is to denote no of other projects who uses this particular packages as dependency., but it is not been referred from UI.

v2/swagger.yaml
arajkumar: This is used in Component Analysis, @deepak1725 will address it as part his spec update PR.

v2/swagger.yaml
arajkumar: Dependency model has following 2 attributes to distinguish public and private vulnerabilities,
 
1. public_vulnerabilities
2. private_vulnerabilities

Both of the above fields are type of `Vulnerability` model. `Vulnerability` contains `PremiumVulnerabilityFields` which will have more information about a vulnerability for a registered user, for unregistered user it will be null.

v2/swagger.yaml
arajkumar: Not used, Legacy stuff, carried over from v1. I will fix it.

arajkumar:
@SriKrishnaPaparaju Incorporated your review comments, PTAL. Thanks.

Swagger hub link: https://app.swaggerhub.com/apis/arajkumar/stack-analyses_api/2.0

v2/stack_analyses.yaml
```diff
@@ -0,0 +1,441 @@
...
+                type: string
+              url:
+                type: string
+      description: Extended Vulnerability details, available only to registered users
+    PublicVulnerability:
+      title: PublicVulnerability
+      description: Publicaly known Vulnerability details with associated CVES
+      type: object
+      properties:
+        cve_ids:
```
deepak1725, line 151: Please refer to this comment https://github.com/fabric8-analytics/fabric8-analytics-server/pull/606#discussion_r401400335

arajkumar: For public vulnerability do have CVE id associated with it.

deepak1725: No, i think Krishna was saying to change it to `vulnerability_id`.

arajkumar: There are 2 fields, id=> snyk id(public and private) and cve_ids => [cve,](only for public).
If you look at the response model, cve_ids field is available only for public vulnerability.

deepak1725: Here Krishna has asked we should not use `cve` word 
PTAL https://github.com/fabric8-analytics/fabric8-analytics-server/pull/609#discussion_r402081401

v2/stack_analyses.yaml
```diff
@@ -0,0 +1,455 @@
...
+  version: "2.0"
+servers:
+  - url: https://f8a-analytics-2445582058137.{environment}.gw.apicast.io/api/v2
+    variables:
+      environment:
+        default: api    # Production server
+        enum:
+          - production # Production server
+          - staging # Staging server
+security:
```
SriKrishnaPaparaju: What does this field referring to ? Please add documentation.

v2/stack_analyses.yaml
```diff
@@ -0,0 +1,455 @@
...
+            - vscode
+      - name: show_transitive
+        in: header
+        description: This is required to enable or disable the transitive support.
+          We need to pass the value as 'true' if transitive is needed. Default value
+          is 'false'.
+        schema:
+          type: boolean
+          default: true
+      requestBody:
```
SriKrishnaPaparaju, line 49: How requests from free tier / registered users are differentiated ?

v2/stack_analyses.yaml
```diff
@@ -0,0 +1,455 @@
...
+                    are mentioned in the README.md
+                  format: binary
+                filePath[]:
+                  type: string
+                  description: The absolute or relative path where the manifest file
+                    is located. This has to be mandatorily provided with the manifest[]
+                    parameter and vice versa.
+        required: true
+      responses:
+        200:
```
SriKrishnaPaparaju, line 70: If UUID is generated would UUID be part of the response while stack analysis is triggered ?

v2/stack_analyses.yaml
```diff
@@ -0,0 +1,455 @@
...
+        401:
+          description: Request Unauthorized
+          content: {}
+  /stack-analyses/{request-id}:
+    get:
+      tags:
+      - Scan Services
+      summary: Get security and other details about an application stack and related
+        recommendations to improve the stack
+      operationId: f8a_scanner.api_v2.stack-analyses-get
```
SriKrishnaPaparaju: f8a reference can be removed. What does this field referring to ?

v2/stack_analyses.yaml
```diff
@@ -0,0 +1,455 @@
...
+          content: {}
+        404:
+          description: Data not found
+          content: {}
+        408:
+          description: Request has timed out on server
+          content: {}
+components:
+  schemas:
+    VulnerabilityForUser:
```
SriKrishnaPaparaju, line 114: this field name is not clear. can we use free-tier & registered-user names ?

v2/stack_analyses.yaml
arajkumar: Once the Auth flow is finalised, most probably we will be sending UUID as part of stack-analyses POST request. UUID will help us in the backend to identify whether the given request is from registered user / un-registered user.

v2/stack_analyses.yaml
arajkumar: No, it won't be part of response.

v2/stack_analyses.yaml
arajkumar: Let me try.

v2/stack_analyses.yaml
arajkumar: This field refers Authorization through 3scale `user_key`. There is no way to add description field for this, I added a comment for more clarity.

arajkumar:
@SriKrishnaPaparaju Incorporated your review comments, PTAL.

v2/stack_analyses.yaml
SriKrishnaPaparaju: UUID creation is something we need to finalize : need to created on the server side. @yzainee ?

v2/stack_analyses.yaml
SriKrishnaPaparaju: If UUID is generated on the server-side (to make sense across all possible integrations), UUID got to be part of the server-side.

v2/stack_analyses.yaml
```diff
@@ -0,0 +1,455 @@
...
+          type: string
+        user_stack_info:
+          $ref: '#/components/schemas/UserStackInfo'
+        manifest_name:
+          type: string
+        recommendation:
+          $ref: '#/components/schemas/StackRecommendation'
+      description: Stack analyses result data
+    StackAnalysesGetResponse:
+      title: StackAnalysesGetResponse
```
SriKrishnaPaparaju, line 431: we need a better name than GetrResponse

v2/stack_analyses.yaml
```diff
@@ -0,0 +1,455 @@
...
+          type: array
+          items:
+            type: object
+            properties: {}
+        alternate:
+          type: array
+          items:
+            $ref: '#/components/schemas/RecommendedPackageData'
+      description: Application stack recommendations
+    StackAnalysesResult:
```
SriKrishnaPaparaju, line 417: This field needs to have two possible responses: Free persona & registered user

v2/stack_analyses.yaml
arajkumar: Fixed it.

v2/stack_analyses.yaml
arajkumar: Actually `analyzed_dependencies` field supposed to have 2 types of results. Just fixed it.

v2/stack_analyses.yaml
SriKrishnaPaparaju: The separate end point (not public facing, internal) for UUID makes sense.

v2/stack_analyses.yaml
```diff
@@ -0,0 +1,469 @@
...
+  schemas:
+    VulnerabilityForRegisteredUser:
+      title: VulnerabilityForRegisteredUser
+      allOf:
+        - $ref: '#/components/schemas/ExtendedVulnerabilityFields'
+        - oneOf:
+          - $ref: '#/components/schemas/PublicVulnerability'
+          - $ref: '#/components/schemas/PrivateVulnerability'
+      description: Vulnerability details for registered users
+    VulnerabilityForFreeTier:
```
SriKrishnaPaparaju, line 122: Please make this plural 'Vulnerabilities'

v2/stack_analyses.yaml
```diff
@@ -0,0 +1,469 @@
...
+    Package:
+      title: Package
+      type: object
+      properties:
+        name:
+          type: string
+        version:
+          type: string
+      description: Package details
+    PackageWithTransitive:
```
SriKrishnaPaparaju, line 209: Please check for renaming with plurals

v2/stack_analyses.yaml
```diff
@@ -0,0 +1,469 @@
...
+      title: PackageDetailsForFreeTier
+      allOf:
+        - $ref: '#/components/schemas/PackageDetailsBase'
+        - type: object
+          properties:
+            public_vulnerabilities:
+              type: array
+              description: Publicly known vulnerability details
+              items:
+                $ref: '#/components/schemas/VulnerabilityForFreeTier'
```
SriKrishnaPaparaju, line 361: Why is this field under public_vulnerabilities ??  Vulnabilities for free tier include complete list & details of public vulnerabilities + count and limited fields for private vulnerabilities.

v2/stack_analyses.yaml
yzainee: For vscode, we would still need it to be generated at the client side. Reasons: while we are doing the processing etc, there are chances where the window might be closed or the ext is closed. In all such cases, it will become very difficult to find a way to send back that uuid to the ext to be saved. To avoid such complexities, its best we generate the uuid at ext and for all other integrations, we do it at the server side.

yzainee: @SriKrishnaPaparaju ^^

v2/stack_analyses.yaml
yzainee: A separate endpoint just to generate uuid at the server side is also fine. We just need to make sure that this endpoint is called at the time of installation and then the uuid is stored locally and is sent for the stack analysis requests. Please dont confuse with my previous comment. These are 2 separate ways of doing the same thing. As long as we dont expect uuid generation at the time of registration (with snyk) i think it should be fine if its generated in the server side via a separate API call.

v2/stack_analyses.yaml
arajkumar: It holds **single** vulnerability details, it is not array of vulnerabilities.

v2/stack_analyses.yaml
arajkumar: It holds single vulnerability details, it is not array of vulnerabilities.

arajkumar: Fixed it.

v2/stack_analyses.yaml
```diff
@@ -138,33 +143,42 @@ components:
...
         exploit:
           type: string
+          example: 
+            - High
+            - Not Defined
         description:
           type: string
+          example: |
+            "## Overview\r\n[`django`](https://pypi.python.org/pypi/django) is a high-level Python Web framework that encourages rapid development and clean, pragmatic design.\r\n\r\nAffected versions of this package expose sensitive information due to not properly restricting the use of a query string that performs certain object filtering. An attacker may obtain sensitive information via a series of requests containing regular expressions, as demonstrated by a `created_by__password__regex` parameter.\r\n\r\n## References\r\n- [Django Vulnerability Description](http://www.djangoproject.com/weblog/2010/dec/22/security/)\r\n- [GitHub Commit](https://github.com/django/django/commit/732198ed5c)\r\n- [Redhat Bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=665373)\r\n- [Openwall](http://www.openwall.com/lists/oss-security/2011/01/03/5)\r\n- [CVE](https://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2010-4534)\r\n"
         fixed_in:
```
yzainee, line 183: the fixed_in field is a list of string. It wont be just a string.

yzainee: example: "initiallyFixedIn": [
        "4.4.5",
        "4.5.2"
      ]

yzainee: If the values to be shown in fixed in field is our computed data, then ignore the comment as we will always show a single recommended version

arajkumar: Nope, it carries snyk's info, I will fix it.

v2/stack_analyses.yaml
SriKrishnaPaparaju: I see that UUID been added as part of the request header.

v2/stack_analyses.yaml
SriKrishnaPaparaju: The response header for the POST call (which triggers the stack analysis) would need send back UUID as part of the response header (if no UUID been sent as part of the request)

v2/stack_analyses.yaml
```diff
@@ -0,0 +1,623 @@
...
+      - Scan Services
+      summary: |
+        Initiates a scan to gather security and other details about 
+        an application stack and related recommendations to improve 
+        the stack
+      parameters:
+      - name: uuid
+        in: header
+        description: Unique identifier for an User.
+        required: false
```
yzainee, line 30: Please make required as true with a default value as NA (similar to what we have in component analysis).
cc: @deepak1725

v2/stack_analyses.yaml
yzainee: uuid field is mandatory in the response header as well. you can check component analysis for example
cc: @deepak1725

PR #609, opened 2020-04-01
Added Component Analyses API Docs

deepak1725:
Adding Component Analyses API Docs
- Updating to OpenAPI version 3
- Adding Separate files for Separate Versions and API Endpoints.
- Adding Interactive API Docs.
- Adding Specs Models for Component Analyses API v2
- Adding Example Response

https://app.swaggerhub.com/apis/deepak1725/analytics-core_api_specs_for_version_2/2

### DIFF
--- a/v2/component_analyses.yaml
+++ b/v2/component_analyses.yaml
@@ -55,7 +55,7 @@ paths:
               examples:
                 RegisteredUser:
                   value:
-                    user_uuid: "b3a51290-79a3-11ea-bb1f-48f17fb770f5"
+                    uuid: "b3a51290-79a3-11ea-bb1f-48f17fb770f5"
                     recommendation:
                       component_analyses:
                         vulnerability:
@@ -74,7 +74,7 @@ paths:
                       vendor_package_link: "https://snyk.io/vuln/maven:org.zeroturnaround%3Azt-zip"
                 FreeUser:
                   value:
-                    user_uuid: "b3a51290-79a3-11ea-bb1f-48f17fb770f5"
+                    uuid: "b3a51290-79a3-11ea-bb1f-48f17fb770f5"
                     recommendation:
                       component_analyses:
                         vulnerability:
@@ -92,7 +92,7 @@ paths:
                       registeration_link: "https://app.snyk.io/login"
                 NoVulnerabilities:
                   value:
-                    user_uuid: "b3a51290-79a3-11ea-bb1f-48f17fb770f5"
+                    uuid: "b3a51290-79a3-11ea-bb1f-48f17fb770f5"
                     recommendation: {}
 
         202:
@@ -128,7 +128,7 @@ components:
       description: Vendor Recommendation information for Premium Response.
       type: object
       properties:
-        user_uuid:
+        uuid:
           type: string
         recommendation:
           type: object
@@ -157,7 +157,7 @@ components:
       description: Vendor Recommendation information for Standard Response.
       type: object
       properties:
-        user_uuid:
+        uuid:
           type: string
         recommendation:
           type: object
@@ -184,7 +184,7 @@ components:
       description: No Vulnerability Response.
       type: object
       properties:
-        user_uuid:
+        uuid:
           type: string
         recommendation:
           type: object
--- a/v2/component_analyses.yaml
+++ b/v2/component_analyses.yaml
@@ -137,7 +137,7 @@ components:
               type: integer
             vendor_package_link:
               type: string
-      description: Vendor Recommendation information for Premium Recommendation.
+          description: Vendor Recommendation information for Premium Recommendation.
 
     FreeUser:
       title: Vendor Public Recommendation Information
--- a/v2/component_analyses.yaml
+++ b/v2/component_analyses.yaml
@@ -99,6 +99,7 @@ components:
   schemas:
     VulnerabilityDetails:
       title: Vulnerability Details
+      description: Information of the vulnerability found
       type: object
       properties:
         vulneability:
@@ -110,10 +111,10 @@ components:
                 type: string
               cvss:
                 type: string
-      description: Information of the vulnerability found
 
     RegisteredUser:
       title: Vendor Private Recommendation Information
+      description: Vendor Recommendation information for Premium Response.
       type: object
       properties:
         recommendation:
@@ -137,10 +138,10 @@ components:
               type: integer
             vendor_package_link:
               type: string
-          description: Vendor Recommendation information for Premium Recommendation.
 
     FreeUser:
       title: Vendor Public Recommendation Information
+      description: Vendor Recommendation information for Standard Response.
       type: object
       properties:
         recommendation:
@@ -160,7 +161,6 @@ components:
               type: integer
             vendor_package_link:
               type: string
-          description: Vendor Recommendation information for Standard Recommendation.
   securitySchemes:
     tokenAuth:
       type: apiKey
--- a/v2/component_analyses.yaml
+++ b/v2/component_analyses.yaml
@@ -34,13 +34,6 @@ paths:
         required: true
         schema:
           type: string
-      - name: security-vendor
-        in: header
-        description: Security Vendor
-        schema:
-          type: string
-          enum:
-          - snyk
       - name: uuid
         in: header
         description: User Identification uuid
--- a/v2/component_analyses.yaml
+++ b/v2/component_analyses.yaml
@@ -86,11 +86,11 @@ components:
           type: string
         severity:
           type: string
-        public_vulnerabilities:
+        public_vulnerabilities_count:
           type: integer
-        private_vulnerabilities:
+        private_vulnerabilities_count:
           type: integer
-        exploitable_vulnerabilities:
+        exploitable_vulnerabilities_count:
           type: integer
         vendor_package_link:
           type: string
@@ -108,7 +108,7 @@ components:
           type: string
         severity:
           type: string
-        public_vulnerabilities:
+        public_vulnerabilities_count:
           type: integer
         vendor_package_link:
           type: string
--- a/v2/component_analyses.yaml
+++ b/v2/component_analyses.yaml
@@ -69,7 +69,7 @@ paths:
                       public_vulnerabilities: 2
                       private_vulnerabilities: 1
                       exploitable_vulnerabilities: 0
-                      snyk_package_link: "https://snyk.io/vuln/maven:org.zeroturnaround%3Azt-zip"
+                      vendor_package_link: "https://snyk.io/vuln/maven:org.zeroturnaround%3Azt-zip"
                 FreeUser:
                   value:
                     recommendation:
@@ -84,7 +84,7 @@ paths:
                       message: "2 vulnerabilities within this dependency"
                       severity: "high"
                       public_vulnerabilities: 2
-                      snyk_package_link: "https://snyk.io/vuln/maven:org.zeroturnaround%3Azt-zip"
+                      vendor_package_link: "https://snyk.io/vuln/maven:org.zeroturnaround%3Azt-zip"
 
         202:
           description: Package is currently Unavailable. Please try again later
--- a/v2/component_analyses.yaml
+++ b/v2/component_analyses.yaml
@@ -51,6 +51,7 @@ paths:
                   oneOf:
                     - $ref: '#/components/schemas/RegisteredUser'
                     - $ref: '#/components/schemas/FreeUser'
+                    - $ref: '#/components/schemas/NoVulnerabilities'
               examples:
                 RegisteredUser:
                   value:
@@ -87,6 +88,9 @@ paths:
                       public_vulnerabilities_count: 2
                       private_vulnerabilities_count: 1
                       registeration_link: "https://app.snyk.io/login"
+                NoVulnerabilities:
+                  value:
+                    recommendation: {}
 
         202:
           description: Package is currently Unavailable. Please try again later
@@ -167,6 +171,15 @@ components:
               type: integer
             registeration_link:
               type: string
+
+    NoVulnerabilities:
+      title: No Vulnerability Response
+      description: No Vulnerability Response.
+      type: object
+      properties:
+        recommendation:
+          type: object
+
   securitySchemes:
     tokenAuth:
       type: apiKey
--- a/v2/component_analyses.yaml
+++ b/v2/component_analyses.yaml
@@ -1,0 +1,312 @@
+openapi: 3.0.1
+info:
+  title: Analytics Core API Specs for Version 2
+  version: "2"
+servers:
+- url: https://f8a-analytics-2445582058137.production.gw.apicast.io/api/v1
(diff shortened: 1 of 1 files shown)


v2/component_analyses.yaml
```diff
@@ -0,0 +1,288 @@
...
+          items:
+            type: integer
+        latest_version:
+          type: array
+          items:
+            type: integer
+        gh_open_issues_count:
+          type: array
+          items:
+            type: integer
```
arajkumar, line 192: I still see lots of unused/unnecessary fields in CA response, can you strip those things as well?

v2/component_analyses.yaml
```diff
@@ -0,0 +1,288 @@
+openapi: 3.0.1
+info:
+  title: Analytics Core API Specs for Version 2
+  version: "2"
+servers:
+- url: https://f8a-analytics-2445582058137.production.gw.apicast.io/api/v2
+  description: "Fabric8 Server"
```
SriKrishnaPaparaju, line 7: Fabric8 is an old brand name. Need to replaced with 'Dependency analytics'

v2/component_analyses.yaml
```diff
@@ -0,0 +1,288 @@
...
+security:
+- tokenAuth: []
```
SriKrishnaPaparaju, line 9: what does this field referring to ? Please add description.

v2/component_analyses.yaml
```diff
@@ -0,0 +1,288 @@
...
+paths:
+  /component-analyses/{eco}/{name}/{ver}:
+    get:
+      tags:
+      - Scan Services
+      summary: Get security and other details about a software component and recommendations
+        to change to a more secured version
+      operationId: f8a_scanner.api_v2.component-analyses
```
SriKrishnaPaparaju, line 17: f8a would need to be replaced. What does this field refer to ?

v2/component_analyses.yaml
```diff
@@ -0,0 +1,288 @@
...
+          type: string
+          enum:
+          - snyk
+      responses:
+        200:
+          description: Detailed response for the requested application component
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ComponentAnalysesResponse'
```
SriKrishnaPaparaju, line 50: Need to model two different responses: free tier and registered user personas

v2/component_analyses.yaml
```diff
@@ -0,0 +1,288 @@
...
+        licenses:
+          type: array
+          items:
+            type: string
+        last_updated:
+          type: array
+          items:
+            type: number
+            format: float
+        vulnerability_ids:
```
SriKrishnaPaparaju, line 91: How do flag a vulnerability as public / private ?

v2/component_analyses.yaml
```diff
@@ -0,0 +1,288 @@
...
+    ComponentAnalysesDataDetails:
+      title: Component Analyses Package and Version
+      type: object
+      properties:
+        version:
+          $ref: '#/components/schemas/Version'
+        package:
+          $ref: '#/components/schemas/Package'
+      description: Component analyses data for package and version
+    CVE:
```
SriKrishnaPaparaju, line 223: what does the field indicate ? CVE == public vulnerability. Lets not use 'CVE' word to indicate public vulnerabilities

v2/component_analyses.yaml
deepak1725: @SriKrishnaPaparaju  This Field is for Authentication. If you can paste the raw code in Swagger Editor, then on Clicking Authorize, we will see something like this..

image

We can't (not allowed) to add a description here. 
So, `description` is added in tokenAuth definition.

v2/component_analyses.yaml
deepak1725: Ack!

v2/component_analyses.yaml
deepak1725: operationId is an optional unique string used to identify an operation. If provided, these IDs must be unique among all operations described in your API.

 Some code generators use this value to name the corresponding methods in code.
    Links can refer to the linked operations by operationId.

Source : https://swagger.io/docs/specification/paths-and-operations/

v2/component_analyses.yaml
deepak1725: Ack!

v2/component_analyses.yaml
deepak1725: Replaced all references to CVE

v2/component_analyses.yaml
deepak1725: Added different Models each for Private and Public Vulnerabilities.
So, if user is Registered,  API response will have `private_vulnerabilities:` and `exploitable_vulnerabilities` in response.

v2/component_analyses.yaml
```diff
@@ -0,0 +1,140 @@
...
+      properties:
+        change_to:
+          type: string
+        component-analyses:
+          $ref: '#/components/schemas/VulnerabilityDetails'
+        message:
+          type: string
+        severity:
+          type: string
+        public_vulnerabilities:
```
arajkumar, line 89: if this field denotes count, please be explicit. e.g. public_vulnerabilities_count.
ditto for below 2 fields.

v2/component_analyses.yaml
arajkumar: Still I don't see the notion of Registered/Unregisterd users in model.

deepak1725: @arajkumar I have added `PrivateRecommendation` and `PublicRecommendation` and means obvious.

v2/component_analyses.yaml
deepak1725: Ack!

v2/component_analyses.yaml
deepak1725: Ack!

v2/component_analyses.yaml
```diff
@@ -0,0 +1,140 @@
...
+          type: string
+        component-analyses:
+          $ref: '#/components/schemas/VulnerabilityDetails'
+        message:
+          type: string
+        severity:
+          type: string
+        public_vulnerabilities_count:
+          type: integer
+        vendor_package_link:
```
arajkumar, line 113: Can you change to url? or more meaningful one?

v2/component_analyses.yaml
```diff
@@ -0,0 +1,140 @@
...
+          type: string
+      description: Vendor Recommendation information for Premium Recommendation.
+
+    PublicRecommendation:
+      title: Vendor Public Recommendation Information
+      type: object
+      properties:
+        change_to:
+          type: string
+        component-analyses:
```
arajkumar, line 105: Use uniform naming convention. Here you should be using underscore(_) to separate words.

v2/component_analyses.yaml
```diff
@@ -0,0 +1,140 @@
...
+          $ref: '#/components/schemas/VulnerabilityDetails'
+        message:
```
arajkumar, line 107: I see the `message` field is more tailored towards the need of extension(lsp)., However it is good to have affected version and recommend versions. This would be helpful for other integrations(clair?) which plans to use component analysis API.

v2/component_analyses.yaml
deepak1725: IMHO `vendor_package_link` is more self-explanatory than plain `url`.

v2/component_analyses.yaml
deepak1725: Ack!

deepak1725: Ack!

v2/component_analyses.yaml
deepak1725: Ack!

v2/component_analyses.yaml
SriKrishnaPaparaju: I still see reference to f8a

v2/component_analyses.yaml
SriKrishnaPaparaju: These field names can be confusing (both the personas can see part / complete set of private vuln. information). Please name these fields based on user persona: free-tier/registered user

SriKrishnaPaparaju: Please recheck if the contents under free-tier / registered user response models clearly indicate only the fields that are displayed in the UI.

v2/component_analyses.yaml
deepak1725: Ack!

v2/component_analyses.yaml
deepak1725: Ack!

v2/component_analyses.yaml
```diff
@@ -0,0 +1,169 @@
...
+
+    FreeUser:
+      title: Vendor Public Recommendation Information
+      type: object
+      properties:
+        recommendation:
+          type: object
+          properties:
+            recommended_version:
+              type: integer
```
arajkumar, line 150: recommended_version type must be a string

v2/component_analyses.yaml
```diff
@@ -0,0 +1,169 @@
...
+      description: Information of the vulnerability found
+
+    RegisteredUser:
+      title: Vendor Private Recommendation Information
+      type: object
+      properties:
+        recommendation:
+          type: object
+          properties:
+            recommended_version:
```
arajkumar, line 122: recommended_version type must be a string

v2/component_analyses.yaml
```diff
@@ -0,0 +1,169 @@
...
+              type: integer
+            private_vulnerabilities_count:
+              type: integer
+            exploitable_vulnerabilities_count:
+              type: integer
+            vendor_package_link:
+              type: string
+      description: Vendor Recommendation information for Premium Recommendation.
+
+    FreeUser:
```
arajkumar, line 142: How does a private vulnerability modelled for FreeUser?

v2/component_analyses.yaml
```diff
@@ -0,0 +1,169 @@
...
+                      severity: "high"
+                      public_vulnerabilities_count: 2
+                      private_vulnerabilities_count: 1
+                      exploitable_vulnerabilities_count: 0
+                      vendor_package_link: "https://snyk.io/vuln/maven:org.zeroturnaround%3Azt-zip"
+                FreeUser:
+                  value:
+                    recommendation:
+                      component_analyses:
+                        cve:
```
arajkumar, line 77: I don't see any field named cve in both FreeUser and RegisteredUser

v2/component_analyses.yaml
```diff
@@ -0,0 +1,169 @@
...
+                          - id: "CVE-2018-1002201"
```
arajkumar, line 78: id should have SYNK-IDs,?

v2/component_analyses.yaml
deepak1725: Ack!

v2/component_analyses.yaml
deepak1725: Ack!

v2/component_analyses.yaml
deepak1725: Updated field name, Id remains similar.
Tried with
```
{
	"gremlin": "gremlin=g.V().has('snyk_vuln_id', 'SNYK-PYTHON-DJANGO-40053').valueMap();"
	
}
```

v2/component_analyses.yaml
deepak1725: Ack!

v2/component_analyses.yaml
deepak1725: For Free User, As discussed, Instead of Snyk Package link, we will have a Registration link in place. We will show `private_vulnerabilities_count` , but not an actual vulnerability.
To see Vulnerability, User has to register.

v2/component_analyses.yaml
```diff
@@ -0,0 +1,176 @@
...
+          type: string
+      - name: ver
+        in: path
+        description: component version
+        required: true
+        schema:
+          type: string
+      - name: security-vendor
+        in: header
+        description: Security Vendor
```
yzainee, line 39: As we are planning to have the uuid generation at the server side, we should make sure that the component analysis too sends the uuid in its header. Initially it should send blank, and the same would be generated and sent back as the comp analysis response. The uuid can then be saved locally and sent in the subsequent calls.

If we are having a completely different endpoint (v2) then in that case do we need to pass this value?

deepak1725: Added UUID. 
Regarding `security-vendor`, It is Used to identify vendor in case of Multiple vendors.
This Field is In Sync with Stack Analyses.

arajkumar: `security-vendor` been removed from stack-analyses request header. Custom header is no longer as we have created a new API version to serve Snyk integration.

deepak1725: @arajkumar Are you planning to change Implementations each time a new vendor Integration kicks In..?

arajkumar: Of course if we drastically change the response model like in the current situation. Since we don't know how we are really going to use the field in future, it is better to avoid using `security-vendor`.

deepak1725: Okay, Removing.

v2/component_analyses.yaml
```diff
@@ -0,0 +1,188 @@
...
+      - name: ver
+        in: path
+        description: component version
+        required: true
+        schema:
+          type: string
+      - name: uuid
+        in: header
+        description: User Identification uuid
+        required: false
```
yzainee, line 40: I would rather make is a required field and send some value as NA if the value is not available. This will help us understand clearly that the uuid is not present rather than think whether uuid is missing or maybe header is missing etc.

yzainee: I dont see the uuid in the response. am i missing something?

deepak1725: Ack!

v2/component_analyses.yaml
```diff
@@ -0,0 +1,197 @@
...
+              schema:
+                items:
+                  oneOf:
+                    - $ref: '#/components/schemas/RegisteredUser'
+                    - $ref: '#/components/schemas/FreeUser'
+                    - $ref: '#/components/schemas/NoVulnerabilities'
+              examples:
+                RegisteredUser:
+                  value:
+                    uuid: "b3a51290-79a3-11ea-bb1f-48f17fb770f5"
```
yzainee, line 58: I think in yesterday's call we had discussed that it should be header, even in case of response. Any specific reason why its added to response body?

deepak1725: Ack!

deepak1725:
> @deepak1725 There is an another component-analysis POST API which takes bulk of packages and returns C.A info for all of those. Are we stopped supporting that from V2?

@arajkumar Can we take Incremental changes..? 
If we find its use-case worth, we will definitely bring it to v2.

arajkumar:
@deepak1725 , that should be fine :) 
We are planning to use the API for claircore integration, but we can start with existing simple CA API first.

PR #614, opened 2020-04-28
Stack analyses API v2 changes according to backbone spec

arajkumar:
Incorporate stack analyses level changes from https://github.com/fabric8-analytics/f8a-server-backbone/pull/212

### DIFF
--- a/v2/stack_analyses.yaml
+++ b/v2/stack_analyses.yaml
@@ -615,7 +615,6 @@ components:
         properties:
           registration_link:
             type: string
-            format: url
           analyzed_dependencies:
             type: array
             items:
--- a/v2/stack_analyses.yaml
+++ b/v2/stack_analyses.yaml
@@ -554,8 +554,8 @@ components:
       discriminator:
         propertyName: registration_status
         mapping:
-          registered: StackAnalysesResultForRegisteredUser
-          freetier: StackAnalysesResultForFreeTier
+          registered: '#/components/schemas/StackAnalysesResultForRegisteredUser'
+          freetier: '#/components/schemas/StackAnalysesResultForFreeTier'
       properties:
         registration_status:
           $ref: '#/components/schemas/RegistrationStatus'
--- a/v2/stack_analyses.yaml
+++ b/v2/stack_analyses.yaml
@@ -549,7 +549,6 @@ components:
         - manifest_name
         - ecosystem
         - unknown_dependencies
-        - transitive_count
         - license_analysis
         - dependencies
         - recommendation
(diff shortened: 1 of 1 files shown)


v2/stack_analyses.yaml
```diff
@@ -593,34 +578,12 @@ components:
               version: 2020.1.alpha
             - name: jsonp
               version: 0.0.1
-        analyzed_direct_dependencies_count:
-          type: integer
-          example: 25              
         transitive_count:
```
prashbnair, line 581: Do we need this count?

v2/stack_analyses.yaml
```diff
@@ -631,33 +594,14 @@ components:
       - $ref: '#/components/schemas/StackAnalysesResult'
```
prashbnair, line 594: shouldnt this be packagedetailsforregistereduser?

prashbnair, line 594: why do we have a ref to stackanalysesresult when that already contains this object which is added via discriminator?

v2/stack_analyses.yaml
arajkumar: Nope. Removing it.!

v2/stack_analyses.yaml
arajkumar: Discriminator property just gives a hint of property/field which can be used to detect the type of a model. It doesn't actually includes the base model anyways,

Please check https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/

prashbnair: It would really make sense to include either registered user or free tier in the stack analysis result using the discriminator and oneof.
The way it is currently done is that stack analysis result is composed inside stack analysis result for free tier. 
The result from the api should be one object which is composed differently based on the use case.

arajkumar: There are 2 ways to use discriminator. 

1. Describe it in the BaseModel itself
2. Describe it in the place where Derived models are used.

I've used approach #1. For more details, Please check https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#discriminator-object

v2/stack_analyses.yaml
```diff
@@ -569,10 +559,7 @@ components:
           freetier: StackAnalysesResultForFreeTier
```
prashbnair, line 559: The spelling of registered is incorrect. The reference needs to be be to schema.

arajkumar: Fixed it. Thanks!

prashbnair: The value for the discriminator mapping needs to be a reference to a schema. It should be #/components/schemas/StackAnalysesResultForRegisteredUser

arajkumar: It can simply refer to the name of the schema instead of ref. But to keep a consistency changed it to reference. Thanks!

PR #620, opened 2020-05-04
Add few missing fields in Stack report response model

arajkumar:


v2/stack_analyses.yaml
```diff
@@ -544,6 +546,11 @@ components:
       title: StackAnalysesResult
       type: object
       required:
+        - version
+        - started_at
+        - finished_at
```
dgpatelgit, line 551: Can we rename this as "ended_at", as we use this attributes in the backbone flow. We can unify the naming of attributes. Also started_at looks perfect with ended_at

v2/stack_analyses.yaml
```diff
@@ -557,6 +564,21 @@ components:
           registered: '#/components/schemas/StackAnalysesResultForRegisteredUser'
           freetier: '#/components/schemas/StackAnalysesResultForFreeTier'
       properties:
+        version:
```
prashbnair, line 567: Why do we need this? Isnt it always V2

v2/stack_analyses.yaml
```diff
@@ -557,6 +564,21 @@ components:
...
+          type: string
+          example: v2
+        started_at:
```
prashbnair, line 570: What is purpose of the started and ended time? Is it displayed in UI?

v2/stack_analyses.yaml
```diff
@@ -557,6 +564,21 @@ components:
...
+          type: string
+          example: "2020-04-06 11:56:08.349405"
+        ended_at:
+          type: string
+          example: "2020-04-06 11:58:18.349405"
+        external_request_id:
```
prashbnair, line 576: Isnt this also part of the request?

v2/stack_analyses.yaml
arajkumar: yep., there is no real need right time. @dgpatelgit , Can you elaborate why you need this?

v2/stack_analyses.yaml
arajkumar: The place where you POST the request(vs extension) and the place where you GET the response is different. So it is good to have id as part of the response.

v2/stack_analyses.yaml
arajkumar: removed it.

v2/stack_analyses.yaml
arajkumar: Removed it

v2/stack_analyses.yaml
arajkumar: Removed it

PR #622, opened 2020-05-08
Implementation of stack analyses V2 api

dgpatelgit:
Implementation of Stack Analyses V2 API as per new code structure.

Stack Analyses GET flow is defined in below sequence diagram::
image

Stack Analyses POST flow is defined in below sequence diagram::
image


bayesian/utility/v2/sa_response_builder.py
```diff
@@ -0,0 +1,69 @@
...
+                "ecosystem": stack_task_result.get("ecosystem", ""),
+                "unknown_dependencies": stack_task_result.get("unknown_dependencies", ""),
+                "license_analysis": stack_task_result.get("license_analysis", ""),
+                "recommendation": {
+                    "companion": recommendations.get("companion", []),
+                    "manifest_file_path": recommendations.get("manifest_file_path", ""),
+                    "usage_outliers": recommendations.get("usage_outliers", [])
+                },
+                "registration_link": stack_task_result.get("registration_link", ""),
+                "analyzed_dependencies": stack_task_result.get("analyzed_dependencies", [])
```
arajkumar, line 61: You may need to add only one field that is `recommendation`, no need to explicitly populate rest of the fields because it is already part of the `stack_task_result`

bayesian/api/api_v2.py
```diff
@@ -172,6 +178,90 @@ def get(ecosystem, package, version):
...
+    @staticmethod
+    def post():
+        """Handle /api/v2/stack-analyses POST REST API."""
+        # 1. Read mandatory params
+        manifest = request.files.get("manifest") or None
+        file_path = request.form.get("file_path") or None
+        ecosystem = request.form.get("ecosystem") or None
+
+        current_app.logger.debug("Mandatory params :: manifest: {} file_path: {} "
+                                 "ecosystem: {}".format(manifest, file_path, ecosystem))
```
arajkumar, line 245: Assigning default values can be part of the `StackAnalyses` request validation.

bayesian/api/api_v2.py
```diff
@@ -172,6 +178,90 @@ def get(ecosystem, package, version):
...
+    @staticmethod
+    def get(external_request_id):
+        """Handle /api/v2/stack-analyses GET REST API."""
+        current_app.logger.debug("GET request_id: {}".format(external_request_id))
+
+        # 1. Set stack analyses object
+        sa = StackAnalyses()
+
+        # 2. Read request data for the request
+        db_result = sa.get_request_data(external_request_id)
```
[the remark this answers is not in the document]
dgpatelgit: This is passed only in three GET function only. Rest places we don't need. Still if you feel its better way, I will make the changes.

bayesian/api/api_v2.py
```diff
@@ -172,6 +178,90 @@ def get(ecosystem, package, version):
...
+                    external_request_id)
+                current_app.logger.error(error_message)
+                return {'error': error_message}, 202
+
+        # 6. Some rare case when workers have not updated the result
+        if stack_result == -1 and recm_data == -1:
+            error_message = "Worker result for request ID '{}' doesn't exist yet".format(
+                external_request_id)
+            current_app.logger.error(error_message)
+            raise HTTPError(404, error=error_message)
```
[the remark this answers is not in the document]
dgpatelgit: Its a good split, let me come up with something around this.

bayesian/utility/v2/stack_analyses.py
```diff
@@ -0,0 +1,243 @@
...
+    def __init__(self, ecosystem=None, manifest=None, file_path=None, show_transitive=None):
+        """Initialize params to be used for ."""
+        self.ecosystem = ecosystem
+        self.manifest = manifest
+        self.file_path = file_path
+        self.show_transitive = show_transitive
+
+        # Private members used during post request.
+        self._new_request_id = None
+        self._request_date_str = None
```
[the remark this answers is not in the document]
dgpatelgit: I will check and clean up.

bayesian/utility/v2/sa_response_builder.py
dgpatelgit: I differ here a bit, I feel its good to keep output format as per API spec from this end point and create a by-pass by just coping entire stack_result as it to response. Say in future by any chance we have modified API for backbone / it gives wrong fields, same will not be passed on to client causing more havoc. It will be filtered by this fixed response structure. All I am thinking is to keep output structure of SA V2 as per Spec irrespective of what response I get from backbone. Let me know if you need a discussion on this.

bayesian/api/api_v2.py
dgpatelgit: You mean to create a default values for each parameter should be read from StackAnalyses class?

bayesian/api/api_v2.py
prashbnair, line 229: Does it make sense to send this error message to the user? Also I think the return code should in the 500 range.

bayesian/api/api_v2.py
prashbnair: If they are mandatory can we put an Assert and throw an exception if they are not present

bayesian/api/api_v2.py
dgpatelgit: This is 404, as given request_id is not valid. So wrong input from user. It makes sense to send this kind of error to user, as they can retry with correct input. More over this is same behaviour from V1.

bayesian/utility/v2/sa_response_builder.py
prashbnair: You are always going to depend on what backbone persists. If backbone changes what it persists, then there are no changes required here.  I think this layer should only serve what backbone persists.

bayesian/utility/v2/stack_analyses.py
```diff
@@ -0,0 +1,243 @@
...
+            return {
+                "status": "success",
+                "submitted_at": self._date_str,
+                "id": self._new_request_id
+            }
+
+    def validate_params(self):
+        """Validate mandatory post params of the request."""
+        # Manifest is mandatory and must be a string.
+        if self.manifest is None or not resolved_files_exist(self.manifest.filename):
```
prashbnair, line 78: These can be done before creating an instance of the class. Even if one of them is not present, we do not need to create it.

bayesian/utility/v2/stack_analyses.py
```diff
@@ -0,0 +1,243 @@
...
+            raise HTTPError(400, error=error_message)
+
+        # Below ecosystem map tries to find the map entry for given ecosystem.
+        ecosystem_map = {
+            "node": "npm",
+            "python": "pypi",
+            "java": "maven"
+        }
+        # If given ecosystem is not found in above map, than uses the value passed in the request.
+        self.ecosystem = ecosystem_map.get(self.ecosystem, self.ecosystem)
```
prashbnair, line 108: This map is not required, this is exactly what the util function is doing

bayesian/utility/v2/stack_analyses.py
```diff
@@ -0,0 +1,243 @@
...
+        current_app.logger.info(self._manifest_file_info)
+
+        # Generate unique request id using UUID, also record timestamp in readable form
+        self._new_request_id = str(uuid.uuid4().hex)
+        self._date_str = str(datetime.datetime.now())
+
+    def get_request_data(self, external_new_request_id):
+        """Read request data for given request id from RDS."""
+        db_result = fetch_sa_request(rdb, external_new_request_id)
+        if db_result is None:
```
prashbnair, line 135: This is already handled where this function is called.

bayesian/utility/v2/stack_analyses.py
```diff
@@ -0,0 +1,243 @@
...
+        deps = {}
+        packages = []
+
+        # Dependency finder
+        d = DependencyFinder()
+        deps = d.scan_and_find_dependencies(self.ecosystem, [self._manifest_file_info],
+                                            self.show_transitive)
+
+        # Build package details.
+        for p in deps['result'][0]['details'][0]['_resolved']:
```
prashbnair, line 161: Can we be a little more defensive here? There are a lot of objects that could probably be None.

bayesian/utility/v2/stack_analyses.py
dgpatelgit: This is not happy case, as most of the API are trigger from plugin, I think saving one object creation is not a big deal. Second, thought of keeping all the spec related stuff inside one class, so that it follows OCP rule.

bayesian/utility/v2/stack_analyses.py
dgpatelgit: Can you point me to the function in util, which does this?

bayesian/utility/v2/stack_analyses.py
dgpatelgit: Point taken, I will correct it.

bayesian/utility/v2/stack_analyses.py
dgpatelgit: Sure, will add extra safety net.

bayesian/api/api_v2.py
prashbnair: if it is wrong request id, then it would have been handled in the previous condition as no row would be returned from RDS, this seems like an error on our side.

bayesian/utility/v2/stack_analyses.py
prashbnair: Not required, I misread the code.

bayesian/utility/v2/stack_analyses.py
prashbnair: My point is that the validation needs to be done at the point where the variable was initialised. If it is done there then why do you need it here. The other thing is that only the API layer should ideally know about HTTP Errors. By passing it down to the object you are making this unusable from other layers. If you want to do this please raise an exception from here and generate an appropriate HTTP error code on the api layer.

bayesian/utility/v2/sa_response_builder.py
dgpatelgit: This is not about persist data, API response is about format of the response. I think format of the response should not change, it might create unexpected parsing errors in client. Response format is what defined by API spec, data in that can be from any layer or it can be missing as well. I strongly believe that API response building should be API servers responsibility with respect to API Spec definition and it must not just pass through what comes from backbone services / database.

bayesian/api/api_v2.py
dgpatelgit: Yes we can do that, but I also validated there data type, so instead of doing too many things at API class, I have moved this responsibility to stack analyses class. I am Ok to move all this checks in API class as well, if you insist. Please comment if you need this movement.

bayesian/api/api_v2.py
dgpatelgit: I am also not sure when response will be -1 from database, I have added this check based on V1 flow, I assume that there might be a strong reason behind this.

bayesian/utility/v2/stack_analyses.py
dgpatelgit: Yes, I agree. My new code as moved HTTP error creation in API layer and responder class simply returns code and data.

bayesian/api/api_v2.py
```diff
@@ -214,14 +304,23 @@ def add_resource_no_matter_slashes(resource, route, endpoint=None, defaults=None
...
                                endpoint='get_component_analysis')
-add_resource_no_matter_slashes(SystemVersion, '/system/version')
 
+# Stack analyses routes
+add_resource_no_matter_slashes(StackAnalysesApi,
+                               '/stack-analyses',
+                               endpoint='post_stack_analyses')
+add_resource_no_matter_slashes(StackAnalysesApi,
+                               '/stack-analyses/<external_request_id>',
+                               endpoint='get_stack_analyses')
```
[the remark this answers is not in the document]
dgpatelgit: This endpoints are different, /stack-analyses and other is with parameter 'external_request_id'

bayesian/utility/v2/sa_response_builder.py
```diff
@@ -0,0 +1,69 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Author: Dharmendra G Patel <dhpatel@redhat.com>
+#
+"""Stack analyses API v2 response builder class."""
+
+
+class StackAnalysesResponseBuilder:
+    """Stack analysis response builder for API v2."""
```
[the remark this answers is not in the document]
dgpatelgit: Comment taken, I will try to correct as much as possible.

bayesian/api/api_v2.py
prashbnair: It will be good if you can add an Assert

bayesian/utility/v2/sa_response_builder.py
arajkumar: Backbone's response model has 1 to 1 relationship with API server response model. API server merely acts as proxy to read the data from RDS in this case.

If something changes in the backbone, how would you ensure API server is not broken?

bayesian/api/api_v2.py
arajkumar: I've used `pydantic` to validate & assign request/response default values. It is quite easy to use and doesn't have a learning curve at all. `Pydantic` facilities declarative way to validate input/output data format.

https://github.com/fabric8-analytics/f8a-server-backbone/blob/ed78b2ce41067c29a963a8423260b8f710099b6e/src/v2/models.py#L224

tests/api/test_api_v2.py
```diff
@@ -54,16 +56,16 @@ def test_error_exception(self, accept_json):
 
     def test_error_status_401(self, accept_json, monkeypatch):
         """Test the /error endpoint. Function Redirect."""
-        monkeypatch.setenv("REDIRECT_STATUS", '401')
+        monkeypatch.setenv('REDIRECT_STATUS', '401')
         response = self.client.get(api_route_for('/_error'), headers=accept_json)
         assert response.status_code == 401
-        monkeypatch.delenv("REDIRECT_STATUS", raising=False)
+        monkeypatch.delenv('REDIRECT_STATUS', raising=False)
```
arajkumar, line 62: delenv is not required. pytest makes sure to undo during the test tear-down, which happens after completion of each test.

tests/api/test_api_v2.py
```diff
@@ -128,5 +130,168 @@ def test_get_component_analyses_with_result_not_none(
...
+    def test_sa_get_with_invalid_id(self, _request_data):
+        """Get endpoint with invalid request id to get NONE data from db."""
+        sa = StackAnalysesApi()
+        self.assertRaises(HTTPError, sa.get, 'request_id')
+
+    @patch('bayesian.api.api_v2.RdbAnalyses.get_request_data', return_value={})
+    @patch('bayesian.api.api_v2.RdbAnalyses.get_stack_result', return_value={})
+    @patch('bayesian.api.api_v2.RdbAnalyses.get_recommendation_data', return_value={})
+    @patch('bayesian.api.api_v2.StackAnalysesResponseBuilder.get_response')
+    def test_sa_get_request_success(self, _get_response, _recommendation_data, _stack_result,
```
arajkumar, line 150: I see that excessive mocking in the tests. It makes harder to refactor the code. I suggest to do mocking only for the external component(in this case RDS). So the recommendation is to "Mock across architecturally significant boundaries, but not within those boundaries".

bayesian/api/api_v2.py
```diff
@@ -172,6 +180,113 @@ def get(ecosystem, package, version):
...
+            'content': manifest.read().decode('utf-8')
+        }
+        logger.info(manifest_file_info)
+
+        # 5. Initiate stack analyses object
+        sa = StackAnalyses(None, ecosystem, manifest_file_info, show_transitive)
+
+        # 6. Post request
+        response_status, response_data = sa.post_request()
+        if response_status == 200 or response_status == 202:
```
arajkumar, line 281: Do we have 202 use case for POST?

bayesian/utility/v2/stack_analyses.py
```diff
@@ -0,0 +1,124 @@
...
+            resolved = deps.get('result', [{}])[0].get('details', [{}])[0].get('_resolved', None)
+            if resolved is not None:
+                for p in resolved:
+                    packages.append({
+                        'name': p.get('package', ''),
+                        'version': p.get('version', ''),
+                        'dependencies': p.get('deps', [])
+                    })
+
+            return 200, {'deps': deps, 'packages': packages}
```
arajkumar, line 92: Using http error codes for normal operation would add more confusion. Can we use exceptions to deal with error situations?

bayesian/utility/v2/stack_analyses.py
```diff
@@ -0,0 +1,124 @@
...
+                                                self.show_transitive)
+
+            # Build package details.
+            resolved = deps.get('result', [{}])[0].get('details', [{}])[0].get('_resolved', None)
+            if resolved is not None:
+                for p in resolved:
+                    packages.append({
+                        'name': p.get('package', ''),
+                        'version': p.get('version', ''),
+                        'dependencies': p.get('deps', [])
```
arajkumar, line 89: `deps` will be in the form of `{package:.., version:..}` right? It also has to be transformed to `{name:.., version:..}` format.

dgpatelgit: This is fixed in commit id '6d05d6c'

bayesian/api/api_v2.py
dgpatelgit: No, we don't this, I will correct.

bayesian/api/api_v2.py
```diff
@@ -172,6 +180,113 @@ def get(ecosystem, package, version):
...
+                            '{} is invalid'.format(file_path)
+            logger.error(error_message)
+            raise HTTPError(400, error=error_message)
+
+        # Ecosystem  is mandatory and must be a string.
+        if ecosystem is None or not isinstance(ecosystem, str):
+            error_message = 'Error processing request. Ecosystem is missing / its value {} ' \
+                            'is invalid'.format(ecosystem)
+            logger.error(error_message)
+            raise HTTPError(400, error=error_message)
```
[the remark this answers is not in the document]
dgpatelgit: I will go through pydantic and make required changes.

bayesian/utility/v2/stack_analyses.py
dgpatelgit: This is fixed in commit 'Review comment to HTTPError from base classes is fixed.'

bayesian/utility/v2/stack_analyses.py
dgpatelgit: This is fixed in commit 'Review comment to HTTPError from base classes is fixed.'

bayesian/utility/v2/sa_response_builder.py
```diff
@@ -0,0 +1,69 @@
...
+                    "companion": recommendations.get("companion", []),
+                    "manifest_file_path": recommendations.get("manifest_file_path", ""),
+                    "usage_outliers": recommendations.get("usage_outliers", [])
+                },
+                "registration_link": stack_task_result.get("registration_link", ""),
+                "analyzed_dependencies": stack_task_result.get("analyzed_dependencies", [])
+            }
+        else:
+            response_data = {
+                "error": "Enable to fetch the result for request id '{}'".format(
```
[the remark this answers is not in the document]
dgpatelgit: This is fixed in commit 'Review comment to HTTPError from base classes is fixed.'

bayesian/utility/v2/backbone_server.py
```diff
@@ -0,0 +1,72 @@
...
+            # Post Backbone stack_aggregator call.
+            BackboneServer.session.post(
+                '{}/api/v2/stack_aggregator'.format(BackboneServer.backbone_host),
+                json=body,
+                params=params)
+            return 0
+        except Exception as e:
+            logger.exception('Exception in aggregator api :: {}'.format(e))
+            print('Exception in aggregator api :: {}'.format(e))
+            return -1
```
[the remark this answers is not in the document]
dgpatelgit: This is fixed in commit 'Review comment to HTTPError from base classes is fixed.'

tests/api/test_api_v2.py
dgpatelgit: This is fixed in commit 'Review comment to HTTPError from base classes is fixed.'

bayesian/api/api_v2.py
dgpatelgit: This is fixed in commit 'Review comment to HTTPError from base classes is fixed.'

bayesian/utility/v2/stack_analyses.py
dgpatelgit: This is fixed in commit 'Review comment to HTTPError from base classes is fixed.'

tests/api/test_api_v2.py
dgpatelgit: Fixed in commit 'Added dependecy injection for RDB class from API'

bayesian/api/api_v2.py
```diff
@@ -172,6 +180,113 @@ def get(ecosystem, package, version):
...
+            error_message = 'Invalid request ID {}.'.format(external_request_id)
+            logger.error(error_message)
+            raise HTTPError(404, error=error_message)
+
+        # 2. Read stack results for the request
+        stack_result = RdbAnalyses.get_stack_result(external_request_id)
+
+        # 4. Read recommendation data for the request
+        recm_data = RdbAnalyses.get_recommendation_data(external_request_id)
+
```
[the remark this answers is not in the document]
dgpatelgit: Fixed in commit 'Added dependecy injection for RDB class from API'

bayesian/utility/v2/sa_response_builder.py
```diff
@@ -0,0 +1,110 @@
...
+        stack_task_result = None
+        stack_audit = None
+        recommendations = []
+
+        if self._stack_result is not None and 'task_result' in self._stack_result:
+            stack_task_result = self._stack_result.get('task_result')
+            stack_audit = stack_task_result.get('_audit', {})
+
+        if self._recm_data is not None and 'task_result' in self._recm_data:
+            recommendations = self._recm_data.get('task_result', {}).get('recommendations', [{}])[0]
```
arajkumar, line 61: This None check seems to be redundant because self._stack_result and self._recm_data has been already validated by _is_request_invalid and _is_request_inprogress

bayesian/utility/v2/sa_response_builder.py
```diff
@@ -0,0 +1,110 @@
...
+
+        if stack_task_result is not None:
```
arajkumar, line 63: ditto

bayesian/utility/v2/sa_response_builder.py
arajkumar, line 61: This has been rectified in the backbone, you no longer would see `recommendations` from the recommendation result. `task_result` would have recommendation details directly!

tests/api/test_api_v2.py
arajkumar: I still see excessive mocking. In this test case, it seems to be valid to mock only get_request_data & get_stack_result, nothing else should be mocked.

arajkumar:
> From Getting Multiple Lists, we are now on Getting Single List.

Previously it was a list due to maven legacy requirements, Once we used to send list of pom.xml files which are part of the project to the platform to perform the analysis. However it is no longer used, we compute effective pom at the extension level and only resolved/computed dependency details are passed to the platform. So we no longer need to send list of pom.

> From Reading Ecosystem in Headers, we are now reading it from Body

Custom request Headers mostly used to pass global states to the system(e.g. cookie, auth,..etc). It is not really good to use it to pass parameters to the resource. We could have used query parameters or path based arguments, but in this case we use form-data because it is already been used to carry manifest data., so associating manifest specific details as form data doesn't do any harm.

> filePath[] which was used to be List, is not file_path.
> Is there any specific reason, why these changes are carried out..?

Same reason, read my answer to your first question!

bayesian/api/api_v2.py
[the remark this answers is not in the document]
dgpatelgit: This changes is implemented.

bayesian/api/api_v2.py
dgpatelgit: This is implemented in commit 'Used pydantic to validated post params'.

bayesian/utility/v2/stack_analyses.py
dgpatelgit: Parameter validation is completely changes now in commit 'Used pydantic to validated post params'. It also takes care of above concern.

bayesian/api/api_v2.py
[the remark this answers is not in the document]
dgpatelgit: This is implemented in commit 'Used pydantic to validated post params'

bayesian/__init__.py
```diff
@@ -55,6 +55,10 @@ def create_app(configfile=None):
     app.route('/api')(lambda: redirect(url_for('api_v1.apiendpoints__slashless')))
     # Likewise for base URL, and make that accessible by name
 
+    # Configure CORS, DONE ONLY FOR DEV TESTING, SHOULD NEVER APPEAR IN PRODUCTION.
+    from flask_cors import CORS
+    CORS(app, resources={r"/api/*": {"origins": "*"}})
+
```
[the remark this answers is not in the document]
dgpatelgit: Comment modified.

bayesian/utility/v2/sa_response_builder.py
dgpatelgit: _is_request_valid checks for -1 values only. _is_request_inprogress check if request is in progress based on either of then being None.

bayesian/utility/v2/sa_response_builder.py
dgpatelgit: Same as above.

bayesian/utility/v2/sa_response_builder.py
```diff
@@ -0,0 +1,110 @@
...
+            # If the response is not ready and the timeout period is over, send error 408
+            if request_timed_out(self._db_result):
+                error_message = 'Stack analysis request {} has timed out. Please retry ' \
+                                'with a new analysis.'.format(self.external_request_id)
+                logger.error(error_message)
+                raise HTTPError(408, error_message)
+            else:
+                error_message = 'Analysis for request ID {} is in progress'.format(
+                    self.external_request_id)
+                logger.error(error_message)
```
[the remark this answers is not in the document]
dgpatelgit: Modified to log warning.

bayesian/utility/v2/sa_response_builder.py
```diff
@@ -0,0 +1,110 @@
...
+                'registration_link': stack_task_result.get('registration_link', ''),
+                'analyzed_dependencies': stack_task_result.get('analyzed_dependencies', [])
+            }
+        else:
+            error_message = 'Unable to fetch the result for request id {}'.format(
+                self.external_request_id)
+            logger.exception(error_message)
+            raise HTTPError(500, error_message)
+
+    def _is_request_invalid(self):
```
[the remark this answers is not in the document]
dgpatelgit: Renamed.

bayesian/utility/v2/stack_analyses.py
```diff
@@ -0,0 +1,117 @@
...
+        self.show_transitive = show_transitive
+
+        # Private members used during post request.
+        self._request_date_str = None
+
+    def post_request(self):
+        """Make stack analyses POST request."""
+        # Generate unique request id using UUID, also record timestamp in readable form
+        self.external_request_id = str(uuid.uuid4().hex)
+        self._request_date_str = str(datetime.datetime.now())
```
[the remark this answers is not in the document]
dgpatelgit: Changes done.

dgpatelgit:
Redundant mocking is remove from all new test cases.

bayesian/utility/v2/sa_response_builder.py
dgpatelgit: So, we don't use *recommendation* data at all in API server?

dgpatelgit: As discussed on Slack I have modified the code in latest commit.

arajkumar: @dgpatelgit , So my concern remains valid right? _is_request_inprogress raises exception if either _stack_result or _recm_data is None. Why do we validating for None again the downstream?

arajkumar: if _is_request_inprogress is already raises exception if one the field is None, then why do we need to check for None in the downstream?

bayesian/utility/v2/sa_response_builder.py
```diff
@@ -0,0 +1,106 @@
...
+
+        # If request is inprogress or timeout, it will raise HTTPError with proper message.
+        self._is_request_inprogress()
+
+        # Proceed with building actual response from data.
+        stack_task_result = None
+        stack_audit = None
+        recommendation = {}
+
+        if self._stack_result is not None and 'task_result' in self._stack_result:
```
arajkumar, line 56: Do we have a situation where self._stack_result is not None but it doesn't have `task_result` in it?

dgpatelgit:
Added stack analyses GET sequence diagram.

bayesian/utility/v2/sa_response_builder.py
dgpatelgit: Don't know, it is object return from backbone server, so added this check for extra safety. Do you see any adverse effect of this extra checks?

bayesian/utility/v2/sa_response_builder.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def _is_request_invalid(self):
+        """If request is invalid than it shall raise an exception."""
+        if self._stack_result == -1 and self._recm_data == -1:
+            error_message = 'Worker result for request ID {} does not exist yet'.format(
+                            self.external_request_id)
+            logger.exception(error_message)
+            raise HTTPError(404, error_message)
+
+    def _is_request_inprogress(self):
```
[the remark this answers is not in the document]
dgpatelgit: This is corrected now.

bayesian/utility/v2/sa_response_builder.py
arajkumar: In that case assertion would be the right choice. Having many conditions hinders readability and testability.

bayesian/utility/v2/sa_response_builder.py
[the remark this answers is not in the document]
arajkumar: I still see same old method name.

bayesian/api/api_v2.py
```diff
@@ -172,6 +182,49 @@ def get(ecosystem, package, version):
...
+        except ValidationError as e:
+            # 2. Check of invalid params and raise exception.
+            error_message = 'Validation error(s) in the request.'
+            for error in e.errors():
+                error_message += ' {} => {}.'.format(error['loc'][0], error['msg'])
+            logger.exception(error_message)
+            raise HTTPError(400, error=error_message)
+
+        # 3. Initiate stack analyses object
+        sa = StackAnalyses(None, sa_post_request)
```
arajkumar, line 222: Probably using keyword args would avoid passing None for first param.
`StackAnalyses(params= sa_post_request)`

bayesian/utility/db_gateway.py
```diff
@@ -63,3 +73,54 @@ def get_ca_data_from_graph(cls, ecosystem, package, version, vendor):
...
+        self.manifest = manifest
+        self.deps = deps
+
+    def get_request_data(self):
+        """Read request data for given request id from RDS."""
+        db_result = fetch_sa_request(rdb, self.request_id)
+        if db_result is None:
+            error_message = 'Invalid request ID {}.'.format(self.request_id)
+            logger.exception(error_message)
+            raise HTTPError(404, error_message)
```
prashbnair, line 97: We should not be raising HTTPErrors from this layer. Please move to API layer

bayesian/utility/db_gateway.py
```diff
@@ -63,3 +73,54 @@ def get_ca_data_from_graph(cls, ecosystem, package, version, vendor):
...
+            do_update_stmt = insert_stmt.on_conflict_do_update(
+                index_elements=['id'],
+                set_=dict(dep_snapshot=self.deps)
+            )
+            rdb.session.execute(do_update_stmt)
+            rdb.session.commit()
+        except SQLAlchemyError as e:
+            logger.exception("Error updating log for request {}, exception {}".format(
+                self.request_id, e))
+            raise HTTPError(500, 'Error while saving request {}'.format(self.request_id))
```
prashbnair, line 126: We should not be raising HTTPErrors from this layer. Please move to API layer

bayesian/utility/v2/backbone_server.py
```diff
@@ -0,0 +1,70 @@
...
+                cls.backbone_host, body, params))
+
+            # Post Backbone stack_aggregator call.
+            BackboneServer.session.post(
+                '{}/api/v2/stack_aggregator'.format(BackboneServer.backbone_host),
+                json=body,
+                params=params)
+        except Exception as e:
+            logger.exception('Aggregator api throws exception {}'.format(e))
+            raise HTTPError(500, error="Error while reaching aggregator service")
```
prashbnair, line 54: Same issue here, please correct across the code.

bayesian/utility/db_gateway.py
arajkumar: @prashbnair , I was the one who suggested to use exceptions instead of lots of nested if..else. Do you want him to use custom exception classes instead of throwing generic HttpError from business layer or you don't want to use exceptions at all? 

Unlike any other programming languages, using exceptions are prevalent in Python and even Python core developers recommends to use for flow control. P.S. https://docs.python.org/2/glossary.html#term-eafp

bayesian/utility/db_gateway.py
dgpatelgit: This is explained by Arun in previous comment.

bayesian/utility/v2/backbone_server.py
dgpatelgit: This all was implemented after discussion with Arun.

bayesian/utility/v2/sa_response_builder.py
dgpatelgit: I will revert to only None check for stack and recommendation response.

bayesian/utility/v2/sa_response_builder.py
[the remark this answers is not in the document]
dgpatelgit: My mistake, correcting now to more appropriate name as '_raise_if_inprogress_or_timeout'

bayesian/api/api_v2.py
dgpatelgit: I think now i don't need external_request_id parameter itself for this class now. So, now I have only one parameter as 'params' and its mandatory. So keyword args approach is not required.

bayesian/api/api_v2.py
```diff
@@ -11,35 +11,85 @@
...
+        'status': 408,
+        'message': 'Stack analysis request has timed out. Please retry with a new analysis',
+        'some_description': 'Exception raised when request id is timed out'
+    },
+    'SAInvalidInputException': {
+        'status': 400,
+        'message': 'Invalid dependencies encountered',
+        'some_description': 'Exception raised during parsing manifest file'
+    }
+}
```
arajkumar, line 92: Using `Api(errors)` and Blueprint error handlers `api_v2.errorhandler(..)` seems to be redundant for me, Why do we need to have both in place?

bayesian/utility/db_gateway.py
```diff
@@ -63,3 +72,78 @@ def get_ca_data_from_graph(cls, ecosystem, package, version, vendor):
...
+            do_update_stmt = insert_stmt.on_conflict_do_update(
+                index_elements=['id'],
+                set_=dict(dep_snapshot=self.deps)
+            )
+            rdb.session.execute(do_update_stmt)
+            rdb.session.commit()
+        except SQLAlchemyError as e:
+            logger.exception("Error updating log for request {}, exception {}".format(
+                self.request_id, e))
+            raise RDBSaveException('Error while saving request {}'.format(self.request_id))
```
arajkumar, line 125: Use `raise YourCustomException(...) from e` pattern to preserve actual exception call trace.

bayesian/utility/db_gateway.py
```diff
@@ -63,3 +72,78 @@ def get_ca_data_from_graph(cls, ecosystem, package, version, vendor):
...
+    """Representation of RDB exception.
+
+    Contains details information on exception caused by RDB server.
+    """
+
+    def __init__(self, message):
+        """Call the superclass constructor and set custom message."""
+        super().__init__(self, message)
+        self.message = message
+
```
arajkumar, line 138: having an explicit constructor is not needed, arguments which passed to Exception could be retrieved using `ex.args`

dgpatelgit: This is fixed now.

PR #624, opened 2020-05-26
[APPAI-1202] Removed code related to cve_sources field.

jparsai:
# Description
Removed code related to cve_source as it is not in use anymore.


yzainee:
Makes me wonder why is this API even needed. changes LGTM.

PR #625, opened 2020-05-28
User management API

prashbnair:


Dockerfile
```diff
@@ -1,9 +1,9 @@
 FROM registry.centos.org/centos/centos:7
 
 ENV LANG=en_US.UTF-8 \
-    F8A_WORKER_VERSION=d403113 \
+    F8A_WORKER_VERSION=58d3025 \
     F8A_AUTH_VERSION=5211e23 \
-    F8A_UTILS=f94a04e
+    F8A_UTILS=3bca34e
```
yzainee, line 6: Any specific reasons why utils is updated?

bayesian/api/user_api.py
```diff
@@ -0,0 +1,70 @@
...
+from fabric8a_auth.errors import AuthError
+
+from bayesian.exceptions import HTTPError
+from bayesian.utility import user_utils
+from bayesian.utility.user_utils import UserException
+
+user_api = Blueprint('user_api', __name__, url_prefix='/user')
+
+
+@user_api.route('/readiness')
```
yzainee, line 17: why do we need readiness and liveness probe here? Its for the pod health check.

bayesian/api/user_api.py
```diff
@@ -0,0 +1,70 @@
...
+
+@user_api.route('/<user_id>', methods=['GET'])
+@login_required
+def get_user(user_id):
+    """Endpoint for getting user details."""
+    if not user_id:
+        raise HTTPError(400, "user id should be present")
+
+    user = user_utils.get_user(user_id)
+    return jsonify(user_id=user.user_id, snyk_api_token=user.snyk_api_token)
```
yzainee, line 37: what happens when snyk_api_key is not present? wont user.snyk_api_key throw exception?

bayesian/utility/user_utils.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def get_user(uuid_val):
+    """Get User."""
+    try:
+        query = rdb.session.query(UserDetails).filter(UserDetails.user_id == uuid_val)
+        user = query.one()
+        return user
+    except SQLAlchemyError:
+        rdb.session.rollback()
```
yzainee, line 21: I dont think we need a rollback here

bayesian/utility/user_utils.py
```diff
@@ -0,0 +1,66 @@
...
+        logger.exception(f"Error fetching user with id {uuid_val}")
+        raise UserException("Error fetching user")
+
+
+def create_user(uuid_val, user_source):
+    """Create User."""
+    try:
+        insert_user_stmt = insert(UserDetails).values(user_id=uuid_val, user_source=user_source,
```
yzainee, line 29: I think you should add strftime to add the date in proper format that we want.

bayesian/api/user_api.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+@user_api.route('', methods=['POST'])
+@login_required
+def create_user():
+    """Endpoint for creating user details."""
+    user_uuid = uuid.uuid4()
+    user_utils.create_user(user_uuid, "SNYK")
```
yzainee, line 45: In case of exception while inserting user in rds, we will still be returning the uuid. Is it intended?

bayesian/utility/user_utils.py
```diff
@@ -0,0 +1,66 @@
...
+        rdb.session.rollback()
+        logger.exception("Error creating new user")
+        raise UserException("Error creating user")
+
+
+def update_user(user_id, snyk_api_token):
+    """Update User."""
+    try:
+        update_stmt = update(UserDetails).where(UserDetails.user_id == user_id).values(
+            dict(snyk_api_token=snyk_api_token, updated_date=datetime.datetime.now(),
```
yzainee, line 45: same comment related to date value as mentioned above

bayesian/utility/user_utils.py
```diff
@@ -0,0 +1,66 @@
...
+                 registered_date=datetime.datetime.now()))
+
+        rdb.session.execute(update_stmt)
+        rdb.session.commit()
+        logger.info(f"User updated with id {user_id}")
+    except SQLAlchemyError:
+        rdb.session.rollback()
```
yzainee, line 52: should we add some retry mechanisms here?

Dockerfile
prashbnair: I had to pull in the latest version of worker to get the RDS model. When I pulled that it required some functions that were present in a later version of utils.

bayesian/api/user_api.py
prashbnair: Will remove it.

bayesian/api/user_api.py
prashbnair: No should return None and the field will be empty.

bayesian/api/user_api.py
prashbnair: No, if there is an exception it is not caught and propagated up after which the exception handler will handle it. So UUID will not be returned.

bayesian/utility/user_utils.py
prashbnair: The dates seem to be in the right format, i checked using pgweb. Maybe it is doing some conversion, i will check this.

Dockerfile
yzainee: ok

bayesian/api/user_api.py
yzainee: ok. just test it out to make sure we dont end up in an exception

bayesian/utility/user_utils.py
yzainee: with strftime, you will be able to store it in a particular format. Then later when we want to query using these values, it will be very easy for us.

bayesian/api/user_api.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+@user_api.route('/liveness')
+def liveness():
+    """Handle the /liveness REST API call."""
+    return jsonify({}), 200
+
+
+@user_api.route('/<user_id>', methods=['GET'])
+@login_required
```
arajkumar, line 30: Can you create a API spec for these endpoints?

bayesian/api/user_api.py
```diff
@@ -0,0 +1,70 @@
...
+def get_user(user_id):
```
arajkumar, line 31: Can you create a API spec for these endpoints?

bayesian/utility/user_utils.py
```diff
@@ -0,0 +1,66 @@
...
+def get_user(uuid_val):
+    """Get User."""
+    try:
+        query = rdb.session.query(UserDetails).filter(UserDetails.user_id == uuid_val)
+        user = query.one()
+        return user
+    except SQLAlchemyError:
+        rdb.session.rollback()
+        logger.exception(f"Error fetching user with id {uuid_val}")
+        raise UserException("Error fetching user")
```
arajkumar, line 23: Follow `raise Exception(...) from e` pattern to preserve actual call trace when rethrowing an exception

bayesian/utility/user_utils.py
```diff
@@ -0,0 +1,66 @@
...
+from f8a_worker.models import (UserDetails)
+from bayesian import rdb
+
+logger = logging.getLogger(__file__)
+
+
+def get_user(uuid_val):
+    """Get User."""
+    try:
+        query = rdb.session.query(UserDetails).filter(UserDetails.user_id == uuid_val)
```
arajkumar, line 17: I guess `query` has to be renamed to `result`.?

tests/utility/test_user_utils.py
```diff
@@ -0,0 +1,39 @@
+"""Test Case for User Management Utilities."""
```
arajkumar, line 1: This seems to be redundant as you already test user_utils through user_api.

bayesian/utility/user_utils.py
```diff
@@ -0,0 +1,66 @@
...
+class UserException(Exception):
+    """Exception for all User Management."""
+
+    def __init__(self, message):
+        """Initialize the exception."""
+        self.message = message
+
+    def __str__(self):
+        """Representation of the exception."""
+        return self.message
```
arajkumar, line 66: This is not required, you can access arguments to custom exception through exception_obj.args field.

bayesian/api/user_api.py
prashbnair: Sure, will do

bayesian/utility/user_utils.py
prashbnair: Done

bayesian/utility/user_utils.py
prashbnair: Done

tests/utility/test_user_utils.py
prashbnair: No, I am mocking out the utility in the api tests.

bayesian/utility/user_utils.py
prashbnair: Done

bayesian/utility/user_utils.py
prashbnair: This converts it to a string but the data type in the DB is datetime.  If we convert to string we will loose the benefits that the DB provides with datetime.

yzainee: oh ok. schema is datetime. then we cannot use strftime.

bayesian/api/user_api.py
prashbnair: https://app.swaggerhub.com/apis/prashbnair/User/1.0.0#/

tests/utility/test_user_utils.py
arajkumar: Don't mock utility from api test and mock only external sources(DB).

bayesian/api/user_api.py
```diff
@@ -0,0 +1,58 @@
...
+    """Endpoint for getting user details."""
+    if not user_id:
+        raise HTTPError(400, "user id should be present")
+
+    user = user_utils.get_user(user_id)
+    return jsonify(user_id=user.user_id, snyk_api_token=user.snyk_api_token)
```
arajkumar, line 25: What is the purpose of returning `snyk_api_token` through API? Is it just to check whether the UUID is associated with registered snyk account? If so, can abstract it with `registration_status` instead of `snyk_api_token`?

tests/utility/test_user_utils.py
prashbnair: Isnt unit testing about testing the unit? Thats why we have fixtures and mocks If there was no external source then we would only need to test the API and dont need to write test cases for anything else. That is not correct.

bayesian/api/user_api.py
prashbnair: Yes this can be changed to status

tests/utility/test_user_utils.py
arajkumar: Unit testing doesn't mean we have to have test files for individual classes. We need to test the module/component as a whole by mocking out the external boundaries. If we have unittest per class/per function, it would definitely hinder ability to refactor the code. Excessive mocking cause refactoring hard. There is a very good post on this. Please check: https://medium.com/@Cyrdup/unit-testing-youre-doing-it-wrong-407a07692989

prashbnair: There are several opinions on unit testing. TDD states that every function needs to first have a test case before it is written. It also says that you should test your function without testing the dependencies. This will make sure that your function is doing what is intended.

bayesian/utility/user_utils.py
yzainee: Are we not going to add any retry mech here? If the operation fails due to some timeout etc, should we not have a retry thing for say 3 times before throwing an exception?

bayesian/api/user_api.py
yzainee: what would be the response that will be sent back to the API caller?

tests/utility/test_user_utils.py
arajkumar: AFAIK, TDD focuses on test first approach(i.e. define tests to cover the spec) and write minimal functionality to pass the test and improve upon with further test cases. It never defines what is unit. Unit is subjective and doesn't mean always a class/function.

prashbnair: When you do TDD, you focus on just the function that you are working on. It encourages you to not bother about other dependencies.  Unit testing like I said even earlier has not defined what a unit but there are multiple ways of doing it so my suggestion is go through the test cases that have been written. If you do not find value in them, it can be discussed. If you want to enforce something and it is being followed all across the project I am fine with that also.

bayesian/utility/user_utils.py
prashbnair: Done, with tenacity library from python

PR #626, opened 2020-05-29
Fixed show transitive flag from boolean type to json

dgpatelgit:


bayesian/utility/v2/stack_analyses.py
```diff
@@ -37,6 +37,9 @@ def __init__(self, params):
         """Initialize params to be used for ."""
         self.params = params
 
+        # Convert show_transitive bool to json as required by dependency finder and backbone api
+        self.params.show_transitive = json.dumps(self.params.show_transitive)
```
arajkumar, line 41: Don't change it globally, only `d.scan_and_find_dependencies` needs it as string.

Ideally the following code should have been modified to fix, but this would affect v1 code path as well. So I'm fine with fixing only v2 specific code which actually calls the `DependencyFinder` api.

https://github.com/fabric8-analytics/fabric8-analytics-utils/blob/1f45e6f9156e07a53ce0b99afed9483ec99c64c8/f8a_utils/dependency_finder.py#L12

dgpatelgit: This class is only for V2 and we need to convert this into json object in dependency finder call as well as backbone body params (which is also json object). So converting this at once make sense.

Changing util is right approach, but will impact v1 as you mentioned, so only I avoided that change.

arajkumar: show_transitive is also passed to backbone right?

arajkumar: The type of `show_transitive` is `bool` at model level. So I won't suggest to change it as the way you did it.

PR #630, opened 2020-06-15
[APPAI-1295] Add environment variable to control the flow for unknown dependencies for component analyses

dgpatelgit:
We need a way to control the unknown package API flow during e2e testing. This PR added a new environment variable 'DISABLE_UNKNOWN_PACKAGE_FLOW' to be enabled on staging to avoid the unknown package API flow for component analyses request.

Jira issue: https://issues.redhat.com/browse/APPAI-1295

### DIFF
--- a/bayesian/api/api_v2.py
+++ b/bayesian/api/api_v2.py
@@ -116,8 +116,12 @@ class ComponentAnalysesApi(Resource):
         Component Analyses:
             - If package is Known (exists in GraphDB (Snyk Edge) returns Json formatted response.
             - If package is not Known:
-                - INVOKE_API_WORKERS flag is up: Trigger bayesianApiFlow to fetch Package details
-                - INVOKE_API_WORKERS flag is down: Trigger bayesianFlow to fetch Package details
+                - SKIP_UNKNOWN_PACKAGE_FLOW flag is up: Skips the unknow package and returns 404
+                - SKIP_UNKONWN_PACKAGE_FLOW flag is down: Than checks below condition.
+                    - INVOKE_API_WORKERS flag is up: Trigger bayesianApiFlow to fetch
+                                                     Package details
+                    - INVOKE_API_WORKERS flag is down: Trigger bayesianFlow to fetch
+                                                       Package details
 
         :return:
             JSON Response
@@ -160,6 +164,14 @@ class ComponentAnalysesApi(Resource):
             metrics_payload.update({"status_code": 200, "value": time.time() - st})
             _session.post(url=METRICS_SERVICE_URL + "/api/v1/prometheus", json=metrics_payload)
             return analyses_result
+        elif os.environ.get("SKIP_UNKNOWN_PACKAGE_FLOW", "") == "1":
+            msg = f"No data found for {ecosystem} package {package}/{version} " \
+                   "and skip unknown package flow is enabled"
+
+            metrics_payload.update({"status_code": 404, "value": time.time() - st})
+            _session.post(url=METRICS_SERVICE_URL + "/api/v1/prometheus", json=metrics_payload)
+
+            raise HTTPError(404, msg)
 
         if os.environ.get("INVOKE_API_WORKERS", "") == "1":
             # Trigger the unknown component ingestion.
--- a/bayesian/api/api_v2.py
+++ b/bayesian/api/api_v2.py
@@ -116,12 +116,12 @@ class ComponentAnalysesApi(Resource):
         Component Analyses:
             - If package is Known (exists in GraphDB (Snyk Edge) returns Json formatted response.
             - If package is not Known:
-                - SKIP_UNKNOWN_PACKAGE_FLOW flag is up: Skips the unknow package and returns 404
-                - SKIP_UNKONWN_PACKAGE_FLOW flag is down: Than checks below condition.
-                    - INVOKE_API_WORKERS flag is up: Trigger bayesianApiFlow to fetch
-                                                     Package details
-                    - INVOKE_API_WORKERS flag is down: Trigger bayesianFlow to fetch
-                                                       Package details
+                - DISABLE_UNKNOWN_PACKAGE_FLOW flag is 1: Skips the unknown package and returns 202
+                - DISABLE_UNKONWN_PACKAGE_FLOW flag is 0: Than checks below condition.
+                    - INVOKE_API_WORKERS flag is 1: Trigger bayesianApiFlow to fetch
+                                                    Package details
+                    - INVOKE_API_WORKERS flag is 0: Trigger bayesianFlow to fetch
+                                                    Package details
 
         :return:
             JSON Response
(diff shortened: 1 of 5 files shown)


dgpatelgit:
> How this Env `SKIP_UNKNOWN_PACKAGE_FLOW` is different from `INVOKE_API_WORKERS` ..?

During 'unknown' package testing from e2e test cases, we want to completely avoid the flow for unknown package request from e2e tests. As I understand INVOKE_API_WORKERS flag is to control flow between 'bayesianApiFlow' and 'bayesianFlow', but does not avoid worker flows. I might be wrong here, let me know if something is misunderstood?

deepak1725:
> > How this Env `SKIP_UNKNOWN_PACKAGE_FLOW` is different from `INVOKE_API_WORKERS` ..?
> 
> During 'unknown' package testing from e2e test cases, we want to completely avoid the flow for unknown package request from e2e tests. As I understand INVOKE_API_WORKERS flag is to control flow between 'bayesianApiFlow' and 'bayesianFlow', but does not avoid worker flows. I might be wrong here, let me know if something is misunderstood?


I see No Exclusive Benefit of Introducing a separate Environment Variable. 
Anyway if you resort to using an Environment variable, go with Third value for INVOKE_API_WORKERS=2 (or whatever).

Additionally, How are you planning to modify its values from E2E..?

dgpatelgit:
> > > How this Env `SKIP_UNKNOWN_PACKAGE_FLOW` is different from `INVOKE_API_WORKERS` ..?
> > 
> > 
> > During 'unknown' package testing from e2e test cases, we want to completely avoid the flow for unknown package request from e2e tests. As I understand INVOKE_API_WORKERS flag is to control flow between 'bayesianApiFlow' and 'bayesianFlow', but does not avoid worker flows. I might be wrong here, let me know if something is misunderstood?
> 
> I see No Exclusive Benefit of Introducing a separate Environment Variable.
> Anyway if you resort to using an Environment variable, go with Third value for INVOKE_API_WORKERS=2 (or whatever).
> 
> Additionally, How are you planning to modify its values from E2E..?

I agree to usage of INVOKE_API_WORKERS with a new value, I will make necessary changes.
Secondly, we will not invoke this from E2E, there was agreement during e2e review meeting that e2e will run against staging and staging data will remain fix and also unknown dependency package flow from CA and SA will be skipped. This was part of that to skip CA flow. I have to modify Backend server also, that work is in progress. After this, we will enable this flag on staging deployment and on prod we will disable this. Hope this clarifies?

bayesian/api/api_v2.py
```diff
@@ -160,6 +164,14 @@ def get(ecosystem, package, version):
             metrics_payload.update({"status_code": 200, "value": time.time() - st})
             _session.post(url=METRICS_SERVICE_URL + "/api/v1/prometheus", json=metrics_payload)
             return analyses_result
+        elif os.environ.get("SKIP_UNKNOWN_PACKAGE_FLOW", "") == "1":
+            msg = f"No data found for {ecosystem} package {package}/{version} " \
```
yzainee, line 168: msg is misleading

bayesian/api/api_v2.py
```diff
@@ -160,6 +164,14 @@ def get(ecosystem, package, version):
             metrics_payload.update({"status_code": 200, "value": time.time() - st})
             _session.post(url=METRICS_SERVICE_URL + "/api/v1/prometheus", json=metrics_payload)
             return analyses_result
```
yzainee, line 166: Disable for V1 as well

bayesian/api/api_v2.py
```diff
@@ -160,6 +164,14 @@ def get(ecosystem, package, version):
...
+        elif os.environ.get("SKIP_UNKNOWN_PACKAGE_FLOW", "") == "1":
+            msg = f"No data found for {ecosystem} package {package}/{version} " \
+                   "and skip unknown package flow is enabled"
+
+            metrics_payload.update({"status_code": 404, "value": time.time() - st})
```
yzainee, line 171: adding a 404 here doesnt look correct to me. Also it will be tracked on prometheus. We are purposefully stopping ingestion so i dont think we should besending 404, rather it should be 202

openshift/template.yaml
```diff
@@ -63,6 +63,8 @@ objects:
                 key: sentry_dsn
           - name: INVOKE_API_WORKERS
             value: "1"
+          - name: SKIP_UNKNOWN_PACKAGE_FLOW
+            value: "0"
```
yzainee, line 67: Use a macro. absolute values wont get overridden via saas template

bayesian/api_v1.py
```diff
@@ -286,6 +306,11 @@ def get(ecosystem, package, version):
             _session.post(url=METRICS_SERVICE_URL + "/api/v1/prometheus", json=metrics_payload)
 
             return result
+        elif os.environ.get("DISABLE_UNKNOWN_PACKAGE_FLOW", "") == "1":
+            msg = f"No data found for {ecosystem} package {package}/{version} " \
+                   "ingetion flow skipped as DISABLE_UNKNOWN_PACKAGE_FLOW is enabled"
+
+            raise HTTPError(202, msg)
```
yzainee, line 313: how is 202 an error?

prashbnair, line 313: Why do you want to raise an error? Should not the normal flow continue if we do not want to ingest packages?

dgpatelgit: The flow ends here, there no further action in this case.

bayesian/api_v1.py
```diff
@@ -286,6 +306,11 @@ def get(ecosystem, package, version):
             _session.post(url=METRICS_SERVICE_URL + "/api/v1/prometheus", json=metrics_payload)
 
             return result
+        elif os.environ.get("DISABLE_UNKNOWN_PACKAGE_FLOW", "") == "1":
```
yzainee, line 309: instead of using the env variable, you can have a common function in utils or config and then call it. function can be names appropriately.

bayesian/api_v1.py
dgpatelgit: It was returning 404 as first proposal, there was review comment to have 202 and do nothing as we have disabled unknown package flow.

bayesian/api_v1.py
dgpatelgit: This how other ENV flags were used in this flow, so maintain consistency with current code style.

bayesian/api_v1.py
dgpatelgit: It was returning 404 as first proposal, there was review comment to have 202 and do nothing as we have disabled unknown package flow.

dgpatelgit: The flow ends here, there no further action in this case.

bayesian/api_v1.py
yzainee: earlier it was in one place. now its in 2 places

bayesian/api_v1.py
yzainee: 202 Accepted
The request has been accepted for processing, but the processing has not been completed. The request might or might not be eventually acted upon, and may be disallowed when processing occurs.

how can it be error?

yzainee: @arajkumar @prashbnair why cant we simply return a valid response with 202.. why are we raising an error with 202? this doesnt sound right.

PR #631, opened 2020-06-16
version semantics added

deepak1725:
Added Version Semantics

Jira: https://issues.redhat.com/browse/APPAI-1283

### DIFF
--- a/bayesian/utility/v2/ca_response_builder.py
+++ b/bayesian/utility/v2/ca_response_builder.py
@@ -19,6 +19,8 @@ from urllib.parse import quote
 import logging
 from bayesian.utility.db_gateway import GraphAnalyses
 from bayesian.utils import version_info_tuple, convert_version_to_proper_semantic
+import semantic_version as sv
+
 
 logger = logging.getLogger(__file__)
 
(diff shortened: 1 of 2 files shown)


bayesian/utility/v2/ca_response_builder.py
```diff
@@ -62,8 +79,9 @@ def get_component_analyses_response(self):
         """
         logger.info('Executing Vendor Specific Analyses')
         try:
+            version = self.version_formatting(convert_version_to_proper_semantic(self.version))
```
prashbnair, line 82: I am not sure if you have fixed the problem. The problem is that we need to be able to search the graph if the version is not complete. i.e if the user has 2.8.0 in the requirements.txt,  that version should be converted to 2.8. so that the data can be returned. It seems like we are doing this in the response. 
Can you also write a test case of that method?

deepak1725: > It seems like we are doing this in the response.

Please see this, you may get a better Idea.
https://github.com/fabric8-analytics/fabric8-analytics-server/blob/ecccaf721d59099ab3fede88d7da49fbe4351aba/bayesian/utility/v2/ca_response_builder.py#L83
So, the fix is totally Inline with the Problem you has described.

I will be adding Unit Tests Shortly.

PR #636, opened 2020-06-22
[APPAI-1324] Incorrect code and message sent in response: SA

dgpatelgit:
Fixed issue with return value of 404 instead of 400.
Removed general exception while querying database, which happens while running server locally during development, but not a real scenario on staging / production.

### DIFF
--- a/bayesian/utility/db_gateway.py
+++ b/bayesian/utility/db_gateway.py
@@ -88,18 +88,13 @@ class RdbAnalyses:
 
     def get_request_data(self):
         """Read request data for given request id from RDS."""
-        try:
-            db_result = fetch_sa_request(rdb, self.request_id)
-            if db_result is None:
-                error_message = 'Invalid request ID {}.'.format(self.request_id)
-                logger.exception(error_message)
-                raise RDBInvalidRequestException(error_message)
-
-            return db_result
-        except Exception as e:
-            error_message = 'DB query failed for request ID {}.'.format(self.request_id)
+        db_result = fetch_sa_request(rdb, self.request_id)
+        if db_result is None:
+            error_message = 'Invalid request ID {}.'.format(self.request_id)
             logger.exception(error_message)
-            raise RDBInvalidRequestException(error_message) from e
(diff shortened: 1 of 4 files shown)


arajkumar:
>Removed general exception while querying database, which happens while running server locally during development, but not a real scenario on staging / production.

@dgpatelgit Can add bit more context about why it can't happen on staging or prod?

dgpatelgit:
> > Removed general exception while querying database, which happens while running server locally during development, but not a real scenario on staging / production.
> 
> @dgpatelgit Can add bit more context about why it can't happen on staging or prod?

First I will explained why it happened on my local setup, during SA V2 development when I was running API server locally, I have to port forward for PostgreSQL service, if this was not on then I was getting AlchamySQL exception, which was handled using that extra code. 

PostgresSQL service not working on stag / prod is very very rare (I assume), that was the reason why API V1, we have not handled this error. So, on similar lines I modified the current code, However this was very common during my development on local setup and was not able to differentiate failure, so added this generic exception block to handle any kind of exception from DB query.

If you feel that having this generic exception block is better, what should it return during this exception? (I think it should return 500 [Internal Server Error])

bayesian/utility/db_gateway.py
```diff
@@ -88,18 +88,13 @@ def __init__(self, request_id, submit_time=None, manifest=None, deps=None):
...
         """Read request data for given request id from RDS."""
-        try:
-            db_result = fetch_sa_request(rdb, self.request_id)
-            if db_result is None:
-                error_message = 'Invalid request ID {}.'.format(self.request_id)
-                logger.exception(error_message)
-                raise RDBInvalidRequestException(error_message)
-
-            return db_result
-        except Exception as e:
```
prashbnair, line 99: I am not sure if this fix is going to solve the problem reported in this bug. From what I understand, the message was returned from the exception block. Now the exception block has been removed to the exception is not caught and just thrown.

dgpatelgit:
> lgtm

New commit, I have added a safe check to catch any generic exception from db server, PTAL.

PR #639, opened 2020-06-23
[APPAI-1325] [APPAI-1326] [APPAI-1347] SA Handling error codes for 400 and 405

dgpatelgit:
PUT, DELET, POST, GET for SA V2 end point not throwing proper 405 error are fixed now.

Jira Issues:
1. https://issues.redhat.com/browse/APPAI-1325 
2. https://issues.redhat.com/browse/APPAI-1326
3. https://issues.redhat.com/browse/APPAI-1347

### DIFF
--- a/bayesian/api/api_v2.py
+++ b/bayesian/api/api_v2.py
@@ -44,7 +44,8 @@ from bayesian.utility.v2.sa_response_builder import (StackAnalysesResponseBuilde
 from bayesian.utility.v2.stack_analyses import StackAnalyses, SAInvalidInputException
 from bayesian.utility.v2.sa_models import StackAnalysesPostRequest
 from bayesian.utility.v2.backbone_server import BackboneServerException
-from bayesian.utility.db_gateway import RdbAnalyses, RDBSaveException, RDBInvalidRequestException
+from bayesian.utility.db_gateway import (RdbAnalyses, RDBSaveException,
+                                         RDBInvalidRequestException, RDBServerException)
 
 
 errors = {
@@ -223,7 +224,9 @@ class StackAnalysesApi(Resource):
         except SARBRequestInvalidException as e:
             raise HTTPError(400, e.args[0]) from e
         except RDBInvalidRequestException as e:
-            raise HTTPError(400, e.args[0]) from e
+            raise HTTPError(404, e.args[0]) from e
+        except RDBServerException as e:
+            raise HTTPError(500, e.args[0]) from e
         except SARBRequestInprogressException as e:
             raise HTTPError(202, e.args[0]) from e
         except SARBRequestTimeoutException as e:
--- a/bayesian/api/api_v2.py
+++ b/bayesian/api/api_v2.py
@@ -201,10 +201,10 @@ class ComponentAnalysesApi(Resource):
         raise HTTPError(404, msg)
 
 
-class StackAnalysesGetApi(Resource):
-    """Implements stack analysis REST APIs.
+class StackAnalysesApiWithId(Resource):
+    """Implements stack analysis REST APIs with request id.
 
-    Implements /api/v2/stack-analyses REST APIs for POST and GET calls.
+    Implements /api/v2/stack-analyses/<Request Id> REST APIs for POST and GET calls.
     """
 
     method_decorators = [login_required]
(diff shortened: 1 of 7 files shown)


dgpatelgit:
> Get class has post method.. post class has get method.. i dont really understand the fix here. Please help me understand. Moreover, i dont understand how will this fix for other http methods
> @arajkumar PTAL
> also not able to understand how does it fix APPAI-1325

This solution is picked up from API V1 flow, which I found after testing v1 end point for this issue. However, I will wait for Arun to suggest something better if available. 

Second, question APPAI-1325 of missing id will be fixed by this, when api endpoint /api/v2/stack_analyses/ is fired with GET method, it will invoke StackAnalysesPostApi class with GET method, which will return proper message as "Missing Id" to the user.

yzainee:
@dgpatelgit for 1325, im sorry but still confused. GET without id will call POST and return missing ID??

dgpatelgit:
> @dgpatelgit for 1325, im sorry but still confused. GET without id will call POST and return missing ID??

Flask app we register two resource for stack analyses as below:
add_resource_no_matter_slashes(StackAnalysesPostApi,
                               '/stack-analyses',
                               endpoint='post_stack_analyses')
add_resource_no_matter_slashes(StackAnalysesGetApi,
                               '/stack-analyses/<external_request_id>',
                               endpoint='get_stack_analyses')

Both resource are called for GET and POST method, these endpoints are handled by StackAnalysesGetApi and StackAnalysesPostApi class.

arajkumar:
@yzainee Do we really need to address these in API server? Error seems to be related to 3scale mapping.

yzainee:
@arajkumar whichever place, i dont mind. just that these cases should be handled.
Missing id case should be handled in api server

dgpatelgit:
I think names are confusing, I should have not used StackAnalysesPostApi and GetApi. I will rename these resources as StackAnalyses and StackAnalysesWithId

yzainee:
@dgpatelgit single class with 2 functions @get and @post should be good.. why do we have 2 classes in the first place? in the get function call, just assert to see id is present else send back 400

dgpatelgit:
> @dgpatelgit single class with 2 functions @get and @post should be good.. why do we have 2 classes in the first place? in the get function call, just assert to see id is present else send back 400

We have two end points with below return status codes.
/api/v2/stack_analyses
     [GET] -> 400 : Missing request id.
     [POST] --> Process new request, returns 200 or 400 or other error.

/api/v2/stack_analyses/<request_id>
    [GET] -> with invalid request, 400 BAD Request.
    [GET] -> With valid request, 200 or 202 (in progres) or 408 (timeout)
    [POST] --> 405 : No Method.

Any other end point
/api/v2/stack_analyses/test/XXX will raise 404.

So, how do we take care of all this with one resource class?

yzainee:
@app.route('/registrazione', methods=['GET', 'POST'])  use something like this to prevent any other http methods. as for the missing id, that can be done easily via an assert in get

yzainee:
flask itself will handle these for you

dgpatelgit:
@arajkumar in API server we are using add_resource_no_matter_slashes() to create routes for resource. Can we skip all this and define routes using BluePrint() app variable as suggested by Yusuf above.

dgpatelgit:
Implemented new routes using flask_restful.Api instead of using existing Resource registration flow. However this will still give 404 (NOT FOUND) status for stack-analyses/ with [GET] method. Note '/' at the end, In order to handle this we need to defined a new route as below:

@api_v2.route('/stack-analyses/', methods=['GET'])

which I think is not correct as we don't have GET with stack-analyses/ as API from CRA.
@arajkumar and @yzainee  Please provide your comments on this.

yzainee:
I dont thik we need a separate function for that. 404 is a valid error in this case.

dgpatelgit:
> I dont thik we need a separate function for that. 404 is a valid error in this case.

If this is OK, can we get approval for this?

yzainee:
@dgpatelgit can you provide the dev cluster link which we can test.. It will be easier to verify that way

dgpatelgit:
> @dgpatelgit can you provide the dev cluster link which we can test.. It will be easier to verify that way

Here is dev cluster URL :: http://bayesian-api-dhpatel-fabric8-analytics.devtools-dev.ext.devshift.net/api/v2/stack-analyses

PR #642, opened 2020-06-24
fix: no. of severity count in ca message.

deepak1725:
Problem was we were using Sets which Merges duplicate values and hence we saw only "1 high severity" in message.

Screenshot from 2020 06 23 16 38 30

Fix: Changed Python set to Python list and revisited the logic.

Related to : Jira: https://issues.redhat.com/browse/APPAI-1318

### DIFF
--- a/tests/utility/v2/test_ca_response_builder.py
+++ b/tests/utility/v2/test_ca_response_builder.py
@@ -264,13 +264,6 @@ class ComponentAnalysisResponseBuilderTest(unittest.TestCase):
         count = response_obj.get_total_vulnerabilities()
         self.assertEqual(count, 0)
 
-    def test_get_severity_unknown_values(self):
-        """Test Severity with unknown value."""
-        response_obj = ComponentAnalysisResponseBuilder(self.eco, self.pkg, self.ver)
-        response_obj._cves = [{'severity': ["unknown"]}]
-        severity = response_obj.get_severity()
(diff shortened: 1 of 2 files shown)


deepak1725:
An example value for Stage having 2 high severities:
```
{
	"gremlin": "g.V().has('pecosystem', 'maven').has('pname', 'org.webjars:summernote').has('version', '0.8.10').as('version').in('has_version').dedup().as('package').select('version').coalesce(out('has_snyk_cve').as('cve').select('package','version','cve').by(valueMap()),select('package','version').by(valueMap()));"
}
```

tests/utility/v2/test_ca_response_builder.py
```diff
@@ -264,13 +264,6 @@ def test_get_total_vulnerabilities(self):
         count = response_obj.get_total_vulnerabilities()
         self.assertEqual(count, 0)
 
-    def test_get_severity_unknown_values(self):
```
prashbnair, line 267: Logic is complicated, can you add a test case

deepak1725: Yes, Sure. That's why I tried to explain it in comments. 
will be adding Tests soon

deepak1725:
@prashbnair I have simplified the Logic to much extend. Added Unit Tests. 
Please review.

PR #645, opened 2020-07-02
[APPAI-1349] [APPAI-1351] incorrect manifest file accepted as a valid input: SA

dgpatelgit:
Fixes below issue with invalid input for stack analyses POST request:

1. https://issues.redhat.com/browse/APPAI-1349
2. https://issues.redhat.com/browse/APPAI-1351

### DIFF
--- a/bayesian/api/api_v2.py
+++ b/bayesian/api/api_v2.py
@@ -246,7 +246,7 @@ def stack_analyses():
         # 2. Check of invalid params and raise exception.
         error_message = 'Validation error(s) in the request.'
         for error in e.errors():
-            error_message += ' {} => {}.'.format(error['loc'][0], error['msg'])
+            error_message += ' {}.'.format(error['msg'])
         logger.exception(error_message)
         raise HTTPError(400, error=error_message) from e
 
--- a/bayesian/api/api_v2.py
+++ b/bayesian/api/api_v2.py
@@ -35,8 +35,7 @@ from bayesian.exceptions import HTTPError
 from bayesian.utils import (get_system_version,
                             server_create_component_bookkeeping,
                             server_create_analysis,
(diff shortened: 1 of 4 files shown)


bayesian/api/api_v2.py
```diff
@@ -243,13 +244,20 @@ def stack_analyses():
...
         logger.exception(error_message)
         raise HTTPError(400, error=error_message) from e
 
+    # 2. Check if ecosystem and manifest file name match
+    ecosystem_from_manifest = get_ecosystem_from_manifest(sa_post_request.manifest.filename)
+    if sa_post_request.ecosystem != ecosystem_from_manifest:
+        raise HTTPError(400, 'Error processing request. Manifest {} and ecosystem {} does '
+                             'not match'.format(sa_post_request.manifest.filename,
+                                                sa_post_request.ecosystem))
+
```
arajkumar, line 260: This validation can be part of sa_models right?

dgpatelgit: Here we want to verify value of two parameter with respect to each other. Can you point me how this can be achieved using pydantic? (I tried initially,but could not find working solution)

arajkumar: Try with root validator.

https://pydantic-docs.helpmanual.io/usage/validators/#root-validators

tests/api/test_api_v2.py
```diff
@@ -335,6 +335,21 @@ def test_sa_post_invalid_ecosystem_value(self):
                                     content_type='multipart/form-data')
         self.assertEqual(response.status_code, 400)
 
+    @patch('bayesian.api.api_v2.StackAnalyses.post_request')
```
arajkumar, line 338: You may not need to mock `StackAnalyses.post_request` because the control would never go there.

PR #646, opened 2020-07-03
ca message update

deepak1725:
Update in CA Message.

Jira: https://issues.redhat.com/browse/APPAI-1336 

### DIFF
--- a/bayesian/utility/v2/ca_response_builder.py
+++ b/bayesian/utility/v2/ca_response_builder.py
@@ -114,13 +114,12 @@ class ComponentAnalysisResponseBuilder:
         self.severity = self.get_severity()
         return self.generate_response()
 
-    def get_message(self):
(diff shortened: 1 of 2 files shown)


bayesian/utility/v2/ca_response_builder.py
```diff
@@ -129,29 +128,43 @@ def get_message(self):
             message += f"and {self.pvt_vul} security advisory with {len(self.severity)} " \
```
prashbnair, line 128: for this message, why does it start with "and" . Will not the message be like "Django - 1.1.1 has and"

deepak1725: @prashbnair Looks like you forgot to expand hidden lines.. :) 
Prior to line 129, on L128 `message += f"{self.public_vul} known security vulnerability"`

PR #649, opened 2020-07-07
removed get_version_without_cves method in ca

deepak1725:
Removed `get_version_without_cves` from CA.


yzainee:
@deepak1725 We still need a way to prevent us from a showing an older version in recommendation. Example if user's version is 2.0.0 and it has vuln, and latest non cve is 1.0.0, then we dont send any recommendation. The if condition was doing that in that function.. I will suggest you keep the code as it is.. and take your time to fix it. This is not an urgent thing needed for the release. Can be taken up later. 
cc: @arajkumar @prashbnair

deepak1725:
As discussed. it will be taken up later.

PR #655, opened 2020-08-03
feat: Implement component analyses batch API

deepak1725:
Component Analyses Batch API Implementation.

Jira: https://issues.redhat.com/browse/APPAI-1400

### DIFF
--- a/bayesian/api/api_v2.py
+++ b/bayesian/api/api_v2.py
@@ -32,12 +32,13 @@ from flask_restful import Api, Resource
 from f8a_worker.utils import MavenCoordinates, case_sensitivity_transform
 from fabric8a_auth.auth import login_required, AuthError
 from bayesian.exceptions import HTTPError
+from bayesian.utility.v2.component_analyses import ca_validate_input, \
+    unknown_package_flow, get_package_version, get_ca_batch_response
 from bayesian.utils import (get_system_version,
                             server_create_component_bookkeeping,
                             server_create_analysis,
                             check_for_accepted_ecosystem)
-from bayesian.utility.v2.ca_response_builder import ComponentAnalyses, \
-    CABatchCall, unknown_package_flow, ca_validate_input, get_package_version
+from bayesian.utility.v2.ca_response_builder import ComponentAnalyses
 from bayesian.utility.v2.sa_response_builder import (StackAnalysesResponseBuilder,
                                                      SARBRequestInvalidException,
                                                      SARBRequestInprogressException,
@@ -73,7 +74,6 @@ METRICS_SERVICE_URL = "http://{}:{}".format(
 worker_count = int(os.getenv('FUTURES_SESSION_WORKER_COUNT', '100'))
 _session = FuturesSession(max_workers=worker_count)
 _resource_paths = []
-logger = logging.getLogger(__name__)
 
 
 @api_v2.route('/readiness')
@@ -223,8 +223,7 @@ class ComponentAnalysesApi(Resource):
             packages_list.append({"name": package, "version": version})
 
         # Perform Component Analyses on Vendor specific Graph Edge.
-        analyses_result, unknown_pkgs = CABatchCall(
-            ecosystem, packages_list).get_ca_batch_response()
+        analyses_result, unknown_pkgs = get_ca_batch_response(ecosystem, packages_list)
 
         if unknown_pkgs:
             api_flow: bool = os.environ.get("INVOKE_API_WORKERS", "") == "1"
--- a/bayesian/api/api_v2.py
+++ b/bayesian/api/api_v2.py
@@ -21,7 +21,7 @@ import time
 import urllib
 import logging
 import re
-from typing import Dict, Tuple, List
+from typing import Dict, Tuple
 
 from requests_futures.sessions import FuturesSession
 from collections import namedtuple
@@ -35,7 +35,7 @@ from f8a_worker.utils import MavenCoordinates, case_sensitivity_transform
 from fabric8a_auth.auth import login_required, AuthError
 from bayesian.exceptions import HTTPError
 from bayesian.utility.v2.component_analyses import ca_validate_input, \
-    unknown_package_flow, get_ca_batch_response
+    unknown_package_flow, get_known_unknown_pkgs
 from bayesian.utils import (get_system_version,
                             server_create_component_bookkeeping,
                             server_create_analysis,
@@ -49,7 +49,8 @@ from bayesian.utility.v2.stack_analyses import StackAnalyses, SAInvalidInputExce
 from bayesian.utility.v2.sa_models import StackAnalysesPostRequest
 from bayesian.utility.v2.backbone_server import BackboneServerException
 from bayesian.utility.db_gateway import (RdbAnalyses, RDBSaveException,
-                                         RDBInvalidRequestException, RDBServerException)
+                                         RDBInvalidRequestException,
+                                         RDBServerException, GraphAnalyses)
 from werkzeug.exceptions import BadRequest
 
 
@@ -209,11 +210,11 @@ class ComponentAnalysesApi(Resource):
     def post():
         """Handle the POST REST API call.
 
-        Component Analyses Batch is a 3 Step Process:
+        Component Analyses Batch is 4 Step Process:
         1. Gather and clean Request.
-        2. Query GraphDB, Builds Stacks Recommendation, Get Unknown Packages Set
-            and Trigger componentApiFlow.
-        3. Handle Unknown Packages and Trigger bayesianApiFlow.
+        2. Query GraphDB.
+        3. Build Stack Recommendation and Build Unknown Packages and Trigger componentApiFlow.
+        4. Handle Unknown Packages and Trigger bayesianApiFlow.
         """
         response_template: Tuple = namedtuple("response_template", ["message", "status"])
         ingestion_disabled_msg: str = "No data found for any package in manifest file. " \
@@ -227,20 +228,24 @@ class ComponentAnalysesApi(Resource):
         input_json: Dict = request.get_json()
         ecosystem: str = input_json.get('ecosystem')
         try:
-            packages_list: List = ca_validate_input(input_json, ecosystem)
+            packages_list, normalised_input_pkgs = ca_validate_input(input_json, ecosystem)
         except BadRequest as br:
             logger.error(br)
             raise HTTPError(400, str(br)) from br
 
-        # Step2: Query GraphDB, Build Unknown packages set and Generates Stack Recommendation.
+        # Step2: Query GraphDB,
         try:
-            stack_recommendation, unknown_pkgs = get_ca_batch_response(ecosystem, packages_list)
+            graph_response = GraphAnalyses.get_batch_ca_data(ecosystem, packages_list)
         except Exception as e:
             msg = "Internal Server Exception. Please contact us if problem persists."
             logger.error(e)
             raise HTTPError(400, msg)
 
-        # Step3: Handle Unknown Packages
+        # Step 3: Build Unknown packages and Generates Stack Recommendation.
+        stack_recommendation, unknown_pkgs = get_known_unknown_pkgs(
+            ecosystem, graph_response, normalised_input_pkgs)
+
+        # Step4: Handle Unknown Packages
         if unknown_pkgs:
             if os.environ.get("DISABLE_UNKNOWN_PACKAGE_FLOW") == "1":
                 # Unknown Packages is Present and INGESTION is DISABLED
--- a/bayesian/api/api_v2.py
+++ b/bayesian/api/api_v2.py
@@ -228,7 +228,7 @@ class ComponentAnalysesApi(Resource):
             packages_list = ca_validate_input(input_json, ecosystem)
         except BadRequest as br:
             logger.debug(br)
-            raise HTTPError(400, str(br))
+            raise HTTPError(400, str(br)) from br
 
         # Step2: Query on Vendor specific Graph Edge and Build Unknown packages set.
         try:
--- a/bayesian/api/api_v2.py
+++ b/bayesian/api/api_v2.py
@@ -36,7 +36,8 @@ from bayesian.utils import (get_system_version,
                             server_create_component_bookkeeping,
                             server_create_analysis,
                             check_for_accepted_ecosystem)
-from bayesian.utility.v2.ca_response_builder import ComponentAnalyses
+from bayesian.utility.v2.ca_response_builder import ComponentAnalyses, validate_version, CABatchCall, \
+    unknown_package_flow
 from bayesian.utility.v2.sa_response_builder import (StackAnalysesResponseBuilder,
                                                      SARBRequestInvalidException,
                                                      SARBRequestInprogressException,
@@ -202,6 +203,70 @@ class ComponentAnalysesApi(Resource):
 
         raise HTTPError(404, msg)
 
+    @staticmethod
+    def post():
+        """Handle the POST REST API call."""
+        input_json = request.get_json()
+        if not input_json:
+            raise HTTPError(400, error="Expected JSON request")
+        if not isinstance(input_json, dict):
+            raise HTTPError(400, error="Expected list of dependencies in JSON request")
+
+        ecosystem = input_json.get('ecosystem')
+        if not check_for_accepted_ecosystem(ecosystem):
+            msg = f"Ecosystem {ecosystem} is not supported for this request"
+            raise HTTPError(400, msg)
+
+        packages_list = []
+        response_template = namedtuple("response_template", ["message", "status"])
+
+        for pkg_obj in input_json.get('packageversions'):
+            package = pkg_obj.get('package')
+            version = pkg_obj.get('version')
+
+            if not all([ecosystem, package, version]):
+                raise HTTPError(422, "provide the valid input.")
+
+            if not validate_version(version):
+                return response_template(
+                    {'message': "Package version should not have special characters."}, 400)
+
+            if ecosystem == 'maven':
+                package = MavenCoordinates.normalize_str(package)
+
+            package = case_sensitivity_transform(ecosystem, package)
+
+            packages_list.append({"name": package, "version": version})
+
+        # Perform Component Analyses on Vendor specific Graph Edge.
+        analyses_result, unknown_pkgs = CABatchCall(ecosystem).get_ca_batch_response(packages_list)
+        disable_ingestion = os.environ.get("DISABLE_UNKNOWN_PACKAGE_FLOW", "") == "1"
+
+        if (analyses_result is None) and disable_ingestion:
+            # No Package is known and Ingestion is disabled.
+            msg = f"No data found for any package in manifest file." \
+                  "Ingestion flow skipped as DISABLE_UNKNOWN_PACKAGE_FLOW is enabled"
+
+            return response_template({'message': msg}, 400)
+
+        if unknown_pkgs:
+            api_flow = os.environ.get("INVOKE_API_WORKERS", "") == "1"
+
+            if disable_ingestion:
+                # Unknown Packages is Present and INGESTION is DISABLED
+                msg = f"No data found for any package in manifest file." \
+                      "Ingestion flow skipped as DISABLE_UNKNOWN_PACKAGE_FLOW is enabled"
+
+                return response_template({'message': msg}, 400)
+
+            unknown_package_flow(ecosystem, unknown_pkgs, api_flow)
+            return analyses_result, 202
+
+        for pkg in analyses_result:
+            # Trigger componentApiFlow for each Known Package
+            server_create_component_bookkeeping(ecosystem, pkg['package'], pkg['version'], g.decoded_token)
+        return analyses_result, 200
+
 
 @api_v2.route('/stack-analyses/<external_request_id>', methods=['GET'])
 @login_required
@@ -322,6 +387,10 @@ add_resource_no_matter_slashes(SystemVersion, '/system/version')
 add_resource_no_matter_slashes(ComponentAnalysesApi,
                                '/component-analyses/<ecosystem>/<package>/<version>',
                                endpoint='get_component_analysis')
+
+add_resource_no_matter_slashes(ComponentAnalysesApi,
+                               '/component-analyses/',
+                               endpoint='post_component_analysis')
 
 
 # ERROR HANDLING
--- a/bayesian/api/api_v2.py
+++ b/bayesian/api/api_v2.py
@@ -33,12 +33,12 @@ from f8a_worker.utils import MavenCoordinates, case_sensitivity_transform
 from fabric8a_auth.auth import login_required, AuthError
 from bayesian.exceptions import HTTPError
 from bayesian.utility.v2.component_analyses import ca_validate_input, \
-    unknown_package_flow, get_package_version, get_ca_batch_response, known_package_flow
+    unknown_package_flow, get_ca_batch_response, known_package_flow
 from bayesian.utils import (get_system_version,
                             server_create_component_bookkeeping,
                             server_create_analysis,
                             check_for_accepted_ecosystem)
-from bayesian.utility.v2.ca_response_builder import ComponentAnalyses
+from bayesian.utility.v2.ca_response_builder import ComponentAnalyses, CABatchResponseBuilder
 from bayesian.utility.v2.sa_response_builder import (StackAnalysesResponseBuilder,
                                                      SARBRequestInvalidException,
                                                      SARBRequestInprogressException,
@@ -48,6 +48,8 @@ from bayesian.utility.v2.sa_models import StackAnalysesPostRequest
 from bayesian.utility.v2.backbone_server import BackboneServerException
 from bayesian.utility.db_gateway import (RdbAnalyses, RDBSaveException,
                                          RDBInvalidRequestException, RDBServerException)
+from werkzeug.exceptions import BadRequest
+
 
 logger = logging.getLogger(__name__)
 
@@ -203,42 +205,58 @@ class ComponentAnalysesApi(Resource):
 
     @staticmethod
     def post():
-        """Handle the POST REST API call."""
-        input_json: dict = request.get_json()
+        """Handle the POST REST API call.
 
-        ca_validate_input(input_json)
+        Component Analyses Batch is a 4 Step Process:
+        1. Gather and clean Request.
+        2. Query GraphDB Vendor Specific Edge
+        3. Build Known Package Response and Trigger componentApiFlow
+        4. Handle Unknown Packages and Trigger bayesianApiFlow
+        """
         response_template = namedtuple("response_template", ["message", "status"])
-        ecosystem: str = input_json.get('ecosystem')
-        disable_ingestion: bool = os.environ.get("DISABLE_UNKNOWN_PACKAGE_FLOW", "") == "1"
         ingestion_disabled_msg: str = "No data found for any package in manifest file. " \
                                       "Ingestion flow skipped as DISABLE_UNKNOWN_PACKAGE_FLOW " \
                                       "is enabled"
         no_package_available_msg: str = "No Package in given manifest is available. " \
-                                        "Packages will be available shortly," \
+                                        "Packages will be available shortly, " \
                                         "Please retry after some time."
 
-        packages_list: list = []
-        for pkg_obj in input_json.get('package_versions'):
-            package, version = get_package_version(pkg_obj, ecosystem)
-            packages_list.append({"name": package, "version": version})
+        # Step1: Gather and clean Request
+        input_json: dict = request.get_json()
+        ecosystem: str = input_json.get('ecosystem')
+        try:
+            packages_list = ca_validate_input(input_json, ecosystem)
+        except BadRequest as br:
+            logger.debug(br)
+            raise HTTPError(400, str(br))
 
-        # Perform Component Analyses on Vendor specific Graph Edge.
-        analyses_result, unknown_pkgs = get_ca_batch_response(ecosystem, packages_list)
+        # Step2: Query on Vendor specific Graph Edge and Build Unknown packages set.
+        try:
+            graph_response, unknown_pkgs = get_ca_batch_response(ecosystem, packages_list)
+        except Exception as e:
+            msg = "Internal Server Exception. Please contact us if problem persists."
+            logger.debug(e)
+            raise HTTPError(400, msg)
 
+        # Step 3: Build known packages Response.
+        analyses_result = []
+        for package in graph_response.get('result', {}).get('data'):
+            pkg_recomendation = CABatchResponseBuilder(ecosystem).generate_recommendation(package)
+            analyses_result.append(pkg_recomendation)
+            known_package_flow(
+                ecosystem, pkg_recomendation["package"], pkg_recomendation["version"])
+
+        # Step4: Handle Unknown Packages
         if unknown_pkgs:
-            api_flow: bool = os.environ.get("INVOKE_API_WORKERS", "") == "1"
-
-            if disable_ingestion:
+            if os.environ.get("DISABLE_UNKNOWN_PACKAGE_FLOW") == "1":
                 # Unknown Packages is Present and INGESTION is DISABLED
-                raise HTTPError(400, error=ingestion_disabled_msg)
-
-            unknown_package_flow(ecosystem, unknown_pkgs, api_flow)
-
+                logger.debug(ingestion_disabled_msg)
+                raise HTTPError(400, ingestion_disabled_msg)
+            unknown_package_flow(ecosystem, unknown_pkgs)
             if not analyses_result:
-                # If None of Packages is Known
+                # If No Package is Known, and all are unknown.
+                logger.debug(no_package_available_msg)
                 return response_template({'error': no_package_available_msg}, 202)
-
-        known_package_flow(ecosystem, analyses_result)
 
         return response_template(analyses_result, 200)
 
--- a/bayesian/api/api_v2.py
+++ b/bayesian/api/api_v2.py
@@ -213,6 +213,9 @@ class ComponentAnalysesApi(Resource):
         ingestion_disabled_msg: str = "No data found for any package in manifest file. " \
                                       "Ingestion flow skipped as DISABLE_UNKNOWN_PACKAGE_FLOW " \
                                       "is enabled"
+        no_package_available_msg: str = f"No Package for {ecosystem} is unavailable. " \
+                                        f"The package will be available shortly," \
+                                        f"Please retry after some time."
 
         packages_list: list = []
         for pkg_obj in input_json.get('package_versions'):
@@ -223,10 +226,6 @@ class ComponentAnalysesApi(Resource):
         analyses_result, unknown_pkgs = CABatchCall(
             ecosystem, packages_list).get_ca_batch_response()
 
-        if (analyses_result is None) and disable_ingestion:
-            # No Package is known and Ingestion is disabled.
-            raise HTTPError(400, error=ingestion_disabled_msg)
-
         if unknown_pkgs:
             api_flow: bool = os.environ.get("INVOKE_API_WORKERS", "") == "1"
 
@@ -235,7 +234,10 @@ class ComponentAnalysesApi(Resource):
                 raise HTTPError(400, error=ingestion_disabled_msg)
 
             unknown_package_flow(ecosystem, unknown_pkgs, api_flow)
-            return response_template(analyses_result, 202)
+
+            if not analyses_result:
+                # If None of Packages is Known
+                return response_template({'error': no_package_available_msg}, 202)
 
         for pkg in analyses_result:
             # Trigger componentApiFlow for each Known Package
--- a/bayesian/api/api_v2.py
+++ b/bayesian/api/api_v2.py
@@ -37,7 +37,7 @@ from bayesian.utils import (get_system_version,
                             server_create_analysis,
                             check_for_accepted_ecosystem)
 from bayesian.utility.v2.ca_response_builder import ComponentAnalyses, \
-    validate_version, CABatchCall, unknown_package_flow
+    CABatchCall, unknown_package_flow, ca_validate_input, get_package_version
 from bayesian.utility.v2.sa_response_builder import (StackAnalysesResponseBuilder,
                                                      SARBRequestInvalidException,
                                                      SARBRequestInprogressException,
@@ -47,7 +47,6 @@ from bayesian.utility.v2.sa_models import StackAnalysesPostRequest
 from bayesian.utility.v2.backbone_server import BackboneServerException
 from bayesian.utility.db_gateway import (RdbAnalyses, RDBSaveException,
                                          RDBInvalidRequestException, RDBServerException)
-
 
 logger = logging.getLogger(__name__)
 
@@ -60,7 +59,6 @@ errors = {
 
 api_v2 = Blueprint('api_v2', __name__, url_prefix='/api/v2')
 rest_api_v2 = Api(api_v2, errors=errors)
-
 
 ANALYSIS_ACCESS_COUNT_KEY = 'access_count'
 TOTAL_COUNT_KEY = 'total_count'
@@ -176,7 +174,7 @@ class ComponentAnalysesApi(Resource):
             return analyses_result
         elif os.environ.get("DISABLE_UNKNOWN_PACKAGE_FLOW", "") == "1":
             msg = f"No data found for {ecosystem} package {package}/{version} " \
-                   "ingetion flow skipped as DISABLE_UNKNOWN_PACKAGE_FLOW is enabled"
+                  "ingetion flow skipped as DISABLE_UNKNOWN_PACKAGE_FLOW is enabled"
 
             return response_template({'error': msg}, 202)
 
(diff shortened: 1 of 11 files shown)


bayesian/utility/v2/ca_response_builder.py
```diff
@@ -18,11 +18,51 @@
...
+        self.packages = packages
+
+    @property
+    def all_packages(self) -> List[namedtuple]:
+        """All Packages in Tuple format."""
+        Package = namedtuple("Package", ["name", "version"])
+        pkg_list = [Package(name=pkg['name'],
+                            version=pkg['version'])
+                    for pkg in self.packages]
+        return pkg_list
```
arajkumar, line 63: So the purpose of this class is to transform the List of package_versions dict to List of package_version namedtuple , so that it becomes hashable and usable with Set? In that case I would recommend define it is a function because class is not a good type for this problem.
```suggestion
Package = namedtuple("Package", ["name", "version"])
def normlize_packages(packages: List[Dict]) -> List[Package]:
    return [Package(p['name'], p['version']) for p in packages]
```

bayesian/utility/v2/ca_response_builder.py
```diff
@@ -18,11 +18,51 @@
...
 import logging
 from bayesian.utility.db_gateway import GraphAnalyses
-from bayesian.utils import version_info_tuple, convert_version_to_proper_semantic
+from bayesian.utils import version_info_tuple, convert_version_to_proper_semantic, \
+    server_create_analysis
+from typing import Dict, List, Tuple, Union, Set
+import re
+from collections import namedtuple
+from abc import ABC
+from flask import g
```
arajkumar, line 27: What is the use of this import?

bayesian/utility/v2/ca_response_builder.py
```diff
@@ -50,7 +90,7 @@ def is_package_known(query_result):
             return False
         return True
 
-    def get_component_analyses_response(self):
+    def get_component_analyses_response(self) -> Union[dict, None]:
```
arajkumar, line 93: It returns a single value right? Do you mean it returns None or Dict, In that case Optional[Dict] is the right type.

bayesian/utility/v2/ca_response_builder.py
```diff
@@ -76,8 +116,78 @@ def get_component_analyses_response(self):
...
+            result = []
+            for package in analyzed_dependencies:
+                result.append(CABatchResponseBuilder(
+                    self.ecosystem, package.name, package.version).generate_recommendation(
+                    graph_response))
+            return result, unknown_pkgs
+
+        except Exception as e:
+            logger.info("Error")
+            logger.error(str(e))
```
arajkumar, line 156: Why both info & error?

bayesian/api/api_v2.py
```diff
@@ -36,7 +36,8 @@
                             server_create_component_bookkeeping,
                             server_create_analysis,
                             check_for_accepted_ecosystem)
-from bayesian.utility.v2.ca_response_builder import ComponentAnalyses
+from bayesian.utility.v2.ca_response_builder import ComponentAnalyses, \
+    validate_version, CABatchCall, unknown_package_flow
```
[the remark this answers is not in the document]
deepak1725: suggestions ..? ~~CABatchCall~~ - ?

bayesian/utility/v2/ca_response_builder.py
deepak1725: `g` is used here https://github.com/fabric8-analytics/fabric8-analytics-server/blob/f5c71b2d2e3a07ce3b0f44c053be6f1646d9db89/bayesian/utility/v2/ca_response_builder.py#L46 

I think its some type of UUID

bayesian/api/api_v2.py
```diff
@@ -202,6 +203,72 @@ def get(ecosystem, package, version):
...
+
+        ecosystem = input_json.get('ecosystem')
+        if not check_for_accepted_ecosystem(ecosystem):
+            msg = f"Ecosystem {ecosystem} is not supported for this request"
+            raise HTTPError(400, msg)
+
+        packages_list = []
+        response_template = namedtuple("response_template", ["message", "status"])
+
+        for pkg_obj in input_json.get('packageversions'):
```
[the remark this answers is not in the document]
deepak1725: Ack!

bayesian/api/api_v2.py
```diff
@@ -202,6 +203,72 @@ def get(ecosystem, package, version):
...
+            package = pkg_obj.get('package')
+            version = pkg_obj.get('version')
+
+            if not all([ecosystem, package, version]):
+                raise HTTPError(422, "provide the valid input.")
+
+            if not validate_version(version):
+                return response_template(
```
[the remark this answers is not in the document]
deepak1725: Ack!

bayesian/api/api_v2.py
```diff
@@ -202,6 +203,72 @@ def get(ecosystem, package, version):
...
+
+                return response_template({'message': msg}, 400)
+
+            unknown_package_flow(ecosystem, unknown_pkgs, api_flow)
+            return analyses_result, 202
+
+        for pkg in analyses_result:
+            # Trigger componentApiFlow for each Known Package
+            server_create_component_bookkeeping(
+                ecosystem, pkg['package'], pkg['version'], g.decoded_token)
```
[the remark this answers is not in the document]
deepak1725: Ack! Moved out.!

bayesian/utility/db_gateway.py
```diff
@@ -72,6 +80,32 @@ def get_ca_data_from_graph(cls, ecosystem, package, version, vendor):
...
+                'ecosystem': ecosystem,
+                'packages': packages
+            }
+        }
+        try:
+            started_at = time.time()
+            response = post(url=gremlin_url, data=json.dumps(payload))
+            response.raise_for_status()
+            elapsed_time = time.time() - started_at
+            logger.info(f"It took {elapsed_time} to fetch results from Gremlin.")
```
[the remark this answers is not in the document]
deepak1725: F strings are used everywhere in code. By Using `format` or anything else, We are Breaking consistancy.

bayesian/utility/v2/ca_response_builder.py
```diff
@@ -76,8 +116,78 @@ def get_component_analyses_response(self):
...
+    """Namespace for Component Analyses Batch call."""
+
+    def __init__(self, ecosystem: str, packages: List[Dict]):
+        """For Flows related to Security Vendor Integrations."""
+        self.ecosystem = ecosystem
+        self.packages = packages
+        self.recommendation = dict()
+        self.response_data = dict()
+
+    def get_ca_batch_response(self) -> Union[Tuple, None]:
```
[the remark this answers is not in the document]
deepak1725: Union type; Union[X, Y] means either X or Y. 
https://stackoverflow.com/questions/33945261/how-to-specify-multiple-return-types-using-type-hints

bayesian/utility/db_gateway.py
```diff
@@ -72,6 +80,32 @@ def get_ca_data_from_graph(cls, ecosystem, package, version, vendor):
         logger.info('Gremlin request took %f seconds', (datetime.now() - start).total_seconds())
         return query_result.json()
 
+    @classmethod
+    def get_batch_ca_data(cls, ecosystem: str, packages: list, query_key: str) -> dict:
```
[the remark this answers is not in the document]
deepak1725: Its because just by changing key, we can refer the whole new Query. Same Procedure is followed in Singular Call.

bayesian/utility/v2/ca_response_builder.py
```diff
@@ -76,8 +116,78 @@ def get_component_analyses_response(self):
...
+        :param packages: List of dict of package, version info.
+
+        :returns:
+            - Json Response
+            - None: Exception/Package not Known.
+        """
+        logger.info('Executing CA Batch Vendor Specific Analyses')
+        try:
+            graph_response = GraphAnalyses().get_batch_ca_data(
+                self.ecosystem, self.packages, 'ca_batch')
```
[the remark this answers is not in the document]
deepak1725: https://github.com/fabric8-analytics/fabric8-analytics-server/pull/655#discussion_r465542807

bayesian/utility/v2/ca_response_builder.py
deepak1725: Union type; Union[X, Y] means either X or Y.
https://stackoverflow.com/questions/33945261/how-to-specify-multiple-return-types-using-type-hints

bayesian/utility/v2/ca_response_builder.py
```diff
@@ -76,8 +116,78 @@ def get_component_analyses_response(self):
...
+        Converts GraphDb output packages into list of Normalised Packages
+        :param graph_response: Graph DB Response
+        :return: list of Normalised Packages
+        """
+        db_pkg_list = []
+        for pack_details in graph_response.get('result').get('data'):
+            pkg_name = pack_details.get('package').get('name', [''])[0]
+            pkg_vr = pack_details.get('version').get('version', [''])[0]
+            db_pkg_list.append({"name": pkg_name, "version": pkg_vr})
+        _normalized_packages = NormalizedPackages(db_pkg_list)
```
prashbnair, line 218: The code in NormalizedPackages can be written here

bayesian/utility/v2/ca_response_builder.py
```diff
@@ -18,11 +18,51 @@
...
 
 
+def validate_version(version: str) -> bool:
+    """Version should not contain special Characters."""
+    if re.findall('[!@#$%^&*()]', version):
+        return False
+    return True
+
+
+def unknown_package_flow(ecosystem: str, unknown_pkgs: Set[namedtuple], api_flow: bool) -> bool:
```
prashbnair, line 42: Why is this defined in the response builder?

bayesian/utility/v2/ca_response_builder.py
```diff
@@ -76,8 +116,78 @@ def get_component_analyses_response(self):
...
+        :return: Unknown Packages to Graph
+        """
+        _normalized_packages = NormalizedPackages(self.packages)
+        input_dependencies = set(_normalized_packages.all_packages)
+
+        return input_dependencies.difference(analyzed_dependencies)
+
+
+class AbstractBaseClass(ABC):
+    """Abstract Class for CA Response Builder."""
```
prashbnair, line 236: This is not required, the code should be written in such a way that the batch call should be an iteration on the single call.

bayesian/utility/db_gateway.py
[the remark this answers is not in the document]
arajkumar: What is the use case for this?

bayesian/utility/db_gateway.py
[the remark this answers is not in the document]
arajkumar: It is mandatory to use either `"msg %s"` or `"msg {}"` in logging statements. I just did a quick look & it is consistently followed in the logging statements.

bayesian/utility/v2/ca_response_builder.py
arajkumar: backbone code which calls unknown packages ingestion doesn't seem to be using that parameter. Can you check?
https://github.com/fabric8-analytics/f8a-server-backbone/blob/master/src/v2/stack_aggregator.py#L385

bayesian/utility/v2/ca_response_builder.py
arajkumar: It is acceptable to use Union to represent two real types. But incase if one of the type is None, you should be using Optional.

See this: https://stackoverflow.com/a/51710151/1942688

bayesian/utility/v2/ca_response_builder.py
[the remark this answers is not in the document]
arajkumar: In this case you are have 2 return values for this function, not either of 2. i.e. it is `x, y`, not `x or y`.

bayesian/api/api_v2.py
```diff
@@ -202,6 +203,72 @@ def get(ecosystem, package, version):
...
+        if not isinstance(input_json, dict):
+            raise HTTPError(400, error="Expected list of dependencies in JSON request")
+
+        ecosystem = input_json.get('ecosystem')
+        if not check_for_accepted_ecosystem(ecosystem):
+            msg = f"Ecosystem {ecosystem} is not supported for this request"
+            raise HTTPError(400, msg)
+
+        packages_list = []
+        response_template = namedtuple("response_template", ["message", "status"])
```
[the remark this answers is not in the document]
deepak1725: There is 1 reference for post and get each. I think this should be Okay.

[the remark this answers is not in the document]
arajkumar: Here you are declaring a type called `response_template`. This particular statement/expression is a constant and shouldn't be evaluated again & again for each call.

bayesian/utility/v2/ca_response_builder.py
deepak1725: Actually it looks like only `componentApiflow` need this .
https://github.com/fabric8-analytics/fabric8-analytics-server/blob/736b821da6dd08dcf2b0b8a4c2bc6c4667778e18/bayesian/utils.py#L99 

AFAIK From Backbone, this flow is not triggered.

bayesian/utility/v2/ca_response_builder.py
deepak1725: Btw I think both are acceptable

Ref Answer first line:
> Optional[...] is a shorthand notation for Union[..., None]

Open to change it to `Optional`, if you say so.

bayesian/api/api_v2.py
```diff
@@ -202,6 +203,72 @@ def get(ecosystem, package, version):
...
+
+            if ecosystem == 'maven':
+                package = MavenCoordinates.normalize_str(package)
+
+            package = case_sensitivity_transform(ecosystem, package)
+
+            packages_list.append({"name": package, "version": version})
+
+        # Perform Component Analyses on Vendor specific Graph Edge.
+        analyses_result, unknown_pkgs = CABatchCall(
```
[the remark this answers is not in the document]
deepak1725: @prashbnair i dont think this can be merged with ComponentAnalyses class, Reason:
- Signature of `CABatchCall` and `ComponentAnalyses` is different. CABatchCall accepts `packages:list` and `ecosystem:str` in its constructor whereas `ComponentAnalyses` accepts `eco:str`, `pkg:str` and `version:str` in its constructor.

- No Common methods that are used in each flow

bayesian/api/api_v2.py
```diff
@@ -202,6 +203,72 @@ def get(ecosystem, package, version):
...
+            ecosystem, packages_list).get_ca_batch_response()
+        disable_ingestion: bool = os.environ.get("DISABLE_UNKNOWN_PACKAGE_FLOW", "") == "1"
+
+        if (analyses_result is None) and disable_ingestion:
```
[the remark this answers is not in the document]
deepak1725: Removed. I think we don't need this check in POST.  PTAL

bayesian/utility/db_gateway.py
```diff
@@ -72,6 +80,32 @@ def get_ca_data_from_graph(cls, ecosystem, package, version, vendor):
...
+        """Component Analyses Batch Call."""
+        logger.info('Executing get_batch_ca_data')
+        ca_batch_query = cls.component_analyses_query.get(query_key)
```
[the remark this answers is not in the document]
deepak1725: Ack!

bayesian/utility/v2/ca_response_builder.py
deepak1725: Because it triggers as a Part of CA response.

bayesian/utility/v2/ca_response_builder.py
deepak1725: We have number of new field Changes introduced in the Batch call.

bayesian/utility/v2/ca_response_builder.py
deepak1725: Changed `NormalizedPackages` class to `normlize_packages` method.

bayesian/utility/v2/ca_response_builder.py
deepak1725: Ack!

bayesian/api/api_v2.py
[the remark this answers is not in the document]
deepak1725: Declaring it separately gives advantage to modify each without affecting another.

bayesian/utility/db_gateway.py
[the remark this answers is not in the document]
deepak1725: Yes, it is.

bayesian/utility/v2/ca_response_builder.py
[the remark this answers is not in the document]
deepak1725: `Tuple[List, Set]` doesn't cover Chances for return type: `None, None` which can also be the case.

bayesian/utility/v2/ca_response_builder.py
arajkumar: Optional is the right type for this.

bayesian/utility/v2/ca_response_builder.py
```diff
@@ -76,8 +116,78 @@ def get_component_analyses_response(self):
...
+        :param analyzed_dependencies: Analyses Packages in GraphDB Response
+        :return: Unknown Packages to Graph
+        """
+        _normalized_packages = NormalizedPackages(self.packages)
+        input_dependencies = set(_normalized_packages.all_packages)
+
+        return input_dependencies.difference(analyzed_dependencies)
+
+
+class AbstractBaseClass(ABC):
```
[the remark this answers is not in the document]
deepak1725: Ack!

bayesian/utility/db_gateway.py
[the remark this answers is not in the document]
deepak1725: Ack!

bayesian/utility/v2/ca_response_builder.py
```diff
@@ -76,8 +116,78 @@ def get_component_analyses_response(self):
...
+
+        Build CA Batch Response.
+        Result is fed to ResponseBuilder.
+        :param packages: List of dict of package, version info.
+
+        :returns:
+            - Json Response
+            - None: Exception/Package not Known.
+        """
+        logger.info('Executing CA Batch Vendor Specific Analyses')
```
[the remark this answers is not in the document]
deepak1725: Ack!

bayesian/utility/v2/ca_response_builder.py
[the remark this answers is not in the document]
deepak1725: Ack!

bayesian/utility/db_gateway.py
```diff
@@ -72,6 +80,32 @@ def get_ca_data_from_graph(cls, ecosystem, package, version, vendor):
...
+        ca_batch_query = inspect.cleandoc(ca_batch_query)
```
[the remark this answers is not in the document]
deepak1725: Removed!

bayesian/utility/v2/ca_response_builder.py
deepak1725: Ack!

bayesian/utility/v2/ca_response_builder.py
prashbnair: The problems with the spec are now leaking into the code. My suggestion is to make post call as a collection of the get call in the spec, we can then change everything together. @arajkumar

bayesian/api/api_v2.py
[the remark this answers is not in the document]
prashbnair: We already have a ComponentAnalysis class, we can have 2 methods that handle the single call and the batch call

bayesian/utility/v2/ca_response_builder.py
deepak1725: @prashbnair We spent a considerable amount of time in Discussing Specs, aligning code with that changes.
Now I don't think its time to fall back...even when in near future we have to revisit the same changes again.

Considering from Front end perspective, By Falling back now, We are introducing extra effort steps: 
- Change Front end in line with CA batch changes.
- Then again align them with new CA Batch field changes.

bayesian/utility/v2/ca_response_builder.py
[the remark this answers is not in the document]
arajkumar: Tuple[Optional[List], Optional[Set]] would be the right type then.

bayesian/utility/v2/ca_response_builder.py
```diff
@@ -351,3 +374,87 @@ def get_cve_maps(self):
...
+        Main function to generate recommendation response.
+
+        :return: Json Response
+        """
+        logger.info("Generating Recommendation")
+        self.version = self.get_version(package_graph_response)
+        self.package = self.get_package(package_graph_response)
+        latest_non_cve_versions: List[str] = package_graph_response.get('package', {}).get(
+            'latest_non_cve_version', [])
+        self._cves = list(map(lambda x: x, package_graph_response.get('cve')))
```
prashbnair, line 398: Do you need this lambda, it is not changing the value.

bayesian/utility/v2/component_analyses.py
```diff
@@ -0,0 +1,159 @@
...
+    """
+    logger.debug('Executing CA Batch Vendor Specific Analyses')
+    try:
+        graph_response: Dict = GraphAnalyses().get_batch_ca_data(ecosystem, packages)
+
+        analyzed_dependencies: Set = set(analysed_package_details(graph_response))
+        unknown_pkgs: Set = get_all_unknown_packages(analyzed_dependencies, packages)
+        result: List = [
+            CABatchResponseBuilder(ecosystem).generate_recommendation(package)
+            for package in graph_response.get('result', {}).get('data')
```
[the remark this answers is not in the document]
deepak1725: Ack!

bayesian/utility/v2/component_analyses.py
```diff
@@ -0,0 +1,159 @@
...
+def get_all_unknown_packages(analyzed_dependencies: Set, packages: List) -> Set:
+    """Get all unknowns packages.
+
+    unknown_packages = input_packages - graphdb_output_packages
+    :param analyzed_dependencies: Analyses Packages in GraphDB Response
+    :param packages: Packages List
+
+    :return: Set of Unknown Packages
+    """
+    normalized_packages = normlize_packages(packages)
```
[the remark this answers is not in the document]
deepak1725: @prashbnair `normlize_packages` are called 2 times
1. To Normalize Input Packages into hashable Set
2. To Normalize GraphDB response into hashable Set 
The Diff of 2 gives us Unknown Packages.

bayesian/utility/v2/component_analyses.py
```diff
@@ -0,0 +1,159 @@
...
+    if not input_json:
+        raise HTTPError(400, error="Expected JSON request")
+
+    if not isinstance(input_json, dict):
+        raise HTTPError(400, error="Expected list of dependencies in JSON request")
+
+    ecosystem = input_json.get('ecosystem')
+    if not check_for_accepted_ecosystem(ecosystem):
+        error_msg: str = f"Ecosystem {ecosystem} is not supported for this request"
+        raise HTTPError(400, error=error_msg)
```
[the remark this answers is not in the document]
deepak1725: Ack!

bayesian/utility/db_gateway.py
```diff
@@ -72,6 +82,28 @@ def get_ca_data_from_graph(cls, ecosystem, package, version, vendor):
...
+        }
+        try:
+            started_at = time.time()
+            response = post(url=gremlin_url, data=json.dumps(payload))
+            response.raise_for_status()
+            elapsed_time = time.time() - started_at
+            logger.info(f"It took {elapsed_time} to fetch results from Gremlin.")
+            return response.json()
+        except Exception as e:
+            logger.error(
```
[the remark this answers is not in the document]
deepak1725: Ack!

bayesian/utility/v2/ca_response_builder.py
deepak1725: Ack!

bayesian/api/api_v2.py
```diff
@@ -202,6 +203,63 @@ def get(ecosystem, package, version):
...
+        response_template = namedtuple("response_template", ["message", "status"])
+        ingestion_disabled_msg: str = "No data found for any package in manifest file. " \
+                                      "Ingestion flow skipped as DISABLE_UNKNOWN_PACKAGE_FLOW " \
+                                      "is enabled"
+        no_package_available_msg: str = "No Package in given manifest is available. " \
+                                        "Packages will be available shortly, " \
+                                        "Please retry after some time."
+
+        # Step1: Gather and clean Request
+        input_json: dict = request.get_json()
```
arajkumar, line 225: Keep uniformity in type name. Use Dict all the places instead dict.

bayesian/api/api_v2.py
```diff
@@ -202,6 +203,63 @@ def get(ecosystem, package, version):
...
+        ecosystem: str = input_json.get('ecosystem')
+        try:
+            packages_list = ca_validate_input(input_json, ecosystem)
+        except BadRequest as br:
+            logger.debug(br)
+            raise HTTPError(400, str(br))
```
arajkumar, line 231: Use `raise YourException from e` pattern everywhere.
```suggestion
            raise HTTPError(400, str(br)) from br
```

bayesian/utility/v2/component_analyses.py
```diff
@@ -0,0 +1,159 @@
...
+
+    Converts GraphDb output packages into set of Normalised hashable Packages
+    :param graph_response: Graph DB Response
+    :return: set of hashable Packages
+    """
+    db_pkg_list = []
+    for pack_details in graph_response.get('result').get('data'):
+        pkg_name = pack_details.get('package').get('name', [''])[0]
+        pkg_vr = pack_details.get('version').get('version', [''])[0]
+        db_pkg_list.append({"name": pkg_name, "version": pkg_vr})
```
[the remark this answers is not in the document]
arajkumar: +1

bayesian/utility/v2/component_analyses.py
```diff
@@ -0,0 +1,153 @@
...
+    """Analyses Package Details from GraphDB.
+
+    Converts GraphDb output packages into set of Normalised hashable Packages
+    :param graph_response: Graph DB Response
+    :return: set of hashable Packages
+    """
+    db_pkg_list = []
+    for pack_details in graph_response.get('result').get('data'):
+        pkg_name = pack_details.get('package').get('name', [''])[0]
+        pkg_vr = pack_details.get('version').get('version', [''])[0]
```
arajkumar, line 135: Instead of being permissive, you assert or fail explicitly incase if response doesn't contain `name` & `version` fields.

tests/api/test_api_v2.py
```diff
@@ -161,6 +162,69 @@ def test_get_component_analyses_with_result_not_none(
...
 
 
+@pytest.mark.usefixtures('client_class')
+class TestCAPostApi(unittest.TestCase):
+    """Component Analyses Unit Tests."""
+
+    @patch('bayesian.api.api_v2.known_package_flow')
+    @patch('bayesian.api.api_v2.unknown_package_flow')
+    @patch('bayesian.api.api_v2.get_ca_batch_response')
+    def test_get_component_analyses_post(self, _mock1, _mock2, _mock3):
```
arajkumar, line 172: These kinda testing pattern would hamper the refactoring. 

Test input and expectation remains constant regardless of the function name. I highly encourage to mock only the external resources access(in this case graph DB access) and always stick with endpoint/route handler as a unit test entry point.

bayesian/utility/db_gateway.py
```diff
@@ -45,9 +45,19 @@ class GraphAnalyses:
...
     }
 
+    def __init__(self):
+        """Initialise GraphAnalyses."""
+        self.ca_batch_query = """
+            epv = [];packages.each {g.V().has('pecosystem', ecosystem).has('pname', it.name)
+            .has('version', it.version).as('version', 'cve').select('version').in('has_version')
+            .dedup().as('package').select('package', 'version', 'cve')
+            .by(valueMap()).by(valueMap()).by(out('has_snyk_cve')
+            .valueMap().fold()).fill(epv);};epv;
```
arajkumar, line 58: `ca_batch_query` need not be instance state, can be a class scoped variable or global

bayesian/utility/db_gateway.py
```diff
@@ -45,9 +45,19 @@ class GraphAnalyses:
...
-            """
+            """,
     }
 
+    def __init__(self):
+        """Initialise GraphAnalyses."""
+        self.ca_batch_query = """
+            epv = [];packages.each {g.V().has('pecosystem', ecosystem).has('pname', it.name)
+            .has('version', it.version).as('version', 'cve').select('version').in('has_version')
+            .dedup().as('package').select('package', 'version', 'cve')
```
arajkumar, line 56: For CA use case you may not need to traverse to Package node at all, Version and CVE node has all information.

bayesian/utility/v2/component_analyses.py
```diff
@@ -0,0 +1,153 @@
...
+    """Normalise Packages into hashable."""
+    return [Package(p['name'], p['version']) for p in packages]
+
+
+def unknown_package_flow(ecosystem: str, unknown_pkgs: Set[namedtuple]) -> bool:
+    """Unknown Package flow. Trigger bayesianApiFlow."""
+    api_flow: bool = os.environ.get("INVOKE_API_WORKERS", "") == "1"
+    for pkg in unknown_pkgs:
+        server_create_analysis(ecosystem, pkg.name, pkg.version, user_profile=g.decoded_token,
+                               api_flow=api_flow, force=False, force_graph_sync=True)
```
arajkumar, line 52: This is a sync call & it would affect the response time. Unknown flow should be executed in the background.

tests/api/test_api_v2.py
deepak1725: Disagree.

bayesian/api/api_v2.py
deepak1725: Why..? Whats the Gain?

bayesian/utility/db_gateway.py
yzainee: latest_non_cve_version, latest_version -> both reside at the package level. Not at the version level. I dont see how removing the pkg node would work here

bayesian/api/api_v2.py
deepak1725: To use `Dict` here, we have to do extra import which create an extra Overhead, and for this Specific purpose, I don't see it a worth.

arajkumar: I can' really understand the overhead in importing a type., that is the norm across code who uses python typing. 

I also see that your other changes imports Dict. e.g. bayesian/utility/v2/ca_response_builder.py

bayesian/api/api_v2.py
arajkumar: It retains the original exception call trace, without that you will lose the stack trace which originally raised the exception.

bayesian/utility/db_gateway.py
arajkumar: Agreed., my bad. :)

bayesian/utility/db_gateway.py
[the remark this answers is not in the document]
arajkumar: Can you look at the existing log statements in api server and align it accordingly ? it all uses `"msg %s"` format.

https://github.com/fabric8-analytics/fabric8-analytics-server/search?q=%22logger.info%22&unscoped_q=%22logger.info%22

tests/api/test_api_v2.py
arajkumar: Can you elaborate why do you disagree with this?

deepak1725: There is a difference b/w Unit Test and Integration Tests and what you are suggesting is a part of Integration Testing
https://www.guru99.com/unit-test-vs-integration-test.html

bayesian/api/api_v2.py
deepak1725: Yes, That's true. in `ca_response_builder` typing module is imported because we can do things like `Tuple[Dict]` means defining types of nested objects.

But in this particular case, `dict` should be suffice.

bayesian/api/api_v2.py
deepak1725: Ack!

bayesian/utility/db_gateway.py
deepak1725: Ack!

bayesian/utility/v2/component_analyses.py
deepak1725: That's why parent component, caller function in `api_v2.py` is in try: catch
this can handle all such errors.

bayesian/api/api_v2.py
arajkumar: What is the point of uniformity and usage recommendation? There won't be any measurable impact on importing a type. 
BTW, you are programming in Python, which is a worst language in terms of performance. You should be more focusing on consistency and readability than micro performance implication.

```suggestion
        input_json: Dict[ = request.get_json()
```

tests/api/test_api_v2.py
arajkumar: Unit testing means testing components in isolation. It doesn't limit the component scope to function.

> Unit example
> Let's take an example - you've got an Order, which can have many OrderLines, a ShippingAddress and a Customer.
> 
> Do we have 4 units here, representing each class? It depends, but most likely it may be easier to treat the whole thing as a Unit. You can write a test which test the whole thing through the Order object. The test is never aware of the existence of the ShippingAddress. It's an internal implementation detail of the Order unit.
> 
> A class doesn't usually make a good Unit, it's usually a collection of classes that is interesting.
> 
> This way of defining a Unit gives you tests, that don't need to change whenever the internals change - that's a good thing. You don't want to change tests on every refactoring - in fact it's a smell.
> 
Reference: https://andrzejonsoftware.blogspot.com/2014/04/tdd-and-rails-what-makes-good-unit.html

Also I would recommend to go through https://medium.com/@Cyrdup/unit-testing-youre-doing-it-wrong-407a07692989

bayesian/api/api_v2.py
deepak1725: > What is the point of uniformity and usage recommendation? There won't be any measurable impact on importing a type.
> BTW, you are programming in Python, which is a worst language in terms of performance. You should be more focusing on consistency and readability than micro performance implication.

Your Statement completely Contradicts your Previous Statement. 
https://github.com/fabric8-analytics/fabric8-analytics-server/pull/655#discussion_r465105468 

FYI: In Logging Discussion: You don't favor format consistency and showed concern about micro Performance.

arajkumar: Do you see the point of consistency in logging? Almost 90% of the logging statement used here in API server follows `"msg %s"` format., why you want to differ from it? Isn't a python recommended way of doing it?

arajkumar: > Your Statement completely Contradicts your Previous Statement.
> #655 (comment)
> 
> FYI: In Logging Discussion: You don't favor format consistency and showed concern about micro Performance.
Even your statement contradicts, you are so concerned about the performance of an import, but not ready to change the recommended logging approach.

bayesian/utility/v2/component_analyses.py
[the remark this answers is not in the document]
deepak1725: Ack!

bayesian/api/api_v2.py
deepak1725: Let's consider Micro Performance too..
Ack this suggestion.
Ack Logging Suggestion.
Converted Log.info to log.debug.

bayesian/utility/db_gateway.py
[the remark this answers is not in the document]
deepak1725: Ack!

bayesian/api/api_v2.py
```diff
@@ -202,6 +205,55 @@ def get(ecosystem, package, version):
...
+        if unknown_pkgs:
+            if os.environ.get("DISABLE_UNKNOWN_PACKAGE_FLOW") == "1":
+                # Unknown Packages is Present and INGESTION is DISABLED
+                logger.error(ingestion_disabled_msg)
+                raise HTTPError(400, ingestion_disabled_msg)
+            unknown_package_flow(ecosystem, unknown_pkgs)
+            if not stack_recommendation:
+                # If No Package is Known, and all are unknown.
+                logger.debug(no_package_available_msg)
+                return response_template({'error': no_package_available_msg}, 202)
```
prashbnair, line 253: Any reason for using both response_template and HttpError for error?

bayesian/utility/db_gateway.py
```diff
@@ -72,6 +80,27 @@ def get_ca_data_from_graph(cls, ecosystem, package, version, vendor):
...
+        }
+        try:
+            started_at = time.time()
+            response = post(url=gremlin_url, data=json.dumps(payload))
+            response.raise_for_status()
+            elapsed_time = time.time() - started_at
+            logger.info("It took %s to fetch results from Gremlin." % elapsed_time)
+            return response.json()
+        except Exception as e:
+            logger.exception(e)
```
prashbnair, line 102: Dont you want to raise an exception and prevent further processing if the graph query fails?

bayesian/utility/v2/component_analyses.py
```diff
@@ -0,0 +1,176 @@
...
+    :param graph_response: Graph DB Response
+    :return: set of hashable Packages, stack recommendation
+    """
+    logger.debug('Triggered Analyses Package Details.')
+    db_pkg_list = []
+    stack_recommendations = []
+    for pack_details in graph_response.get('result').get('data'):
+        pkg_name = pack_details.get('package').get('name', [''])[0]
+        pkg_vr = pack_details.get('version').get('version', [''])[0]
+        db_pkg_list.append({"name": pkg_name, "version": pkg_vr})
```
prashbnair, line 148: Do you need this step? You are converting to a list of dicts and then to a list of packages.

deepak1725: Ack!

bayesian/api/api_v2.py
deepak1725: Actually, in our Monitoring Server, which may come in later, HttpError is taken as Exception, though it's really not an Exception, so I kept it as `response_template`

Changed it to HTTPError for now.

bayesian/utility/db_gateway.py
deepak1725: Encapsulated Parent caller in `api_v2.py` in try: catch. The exception here was kind of redundant. Removed.

bayesian/api/api_v2.py
```diff
@@ -202,6 +206,64 @@ def get(ecosystem, package, version):
...
+                                      "is enabled"
+        no_package_available_msg: str = "No Package in given manifest is available. " \
+                                        "Packages will be available shortly, " \
+                                        "Please retry after some time."
+
+        # Step1: Gather and clean Request
+        input_json: Dict = request.get_json()
+        ecosystem: str = input_json.get('ecosystem')
+        try:
+            packages_list, normalised_input_pkgs = ca_validate_input(input_json, ecosystem)
```
prashbnair, line 231: You can move all the code under one try and multiple catch.  That will improve readability of the code.

bayesian/utility/v2/component_analyses.py
deepak1725: That's a really good Suggestion. 
Actually, for Async call, we have to trigger `Event Loop`. We need to figure out whats overhead of that. Need to do some real-time benchmarking to figure out actual Gain. 

This is something that needs R&D, so I will be putting this into Next Iteration. Thanks

bayesian/api/api_v2.py
deepak1725: Ack!

bayesian/utility/db_gateway.py
```diff
@@ -72,6 +80,24 @@ def get_ca_data_from_graph(cls, ecosystem, package, version, vendor):
...
+            'bindings': {
+                'ecosystem': ecosystem,
+                'packages': packages
+            }
+        }
+        started_at = time.time()
+        response = post(url=gremlin_url, data=json.dumps(payload))
+        response.raise_for_status()
+        elapsed_time = time.time() - started_at
+        logger.info("It took %s to fetch results from Gremlin." % elapsed_time)
```
arajkumar, line 98: This is still a pre formatted string,
```suggestion
        logger.info("It took %s to fetch results from Gremlin.", elapsed_time)
```

bayesian/utility/v2/component_analyses.py
arajkumar: You can consider using Futures instead of Async io similar to backbone request.

bayesian/utility/db_gateway.py
deepak1725: Ack!

bayesian/utility/v2/component_analyses.py
deepak1725: Can you add your suggested changes here.

deepak1725: Thanks, but As stated earlier, I will not be consideting this recommendation now. 

Please let me know if you have any other recommendations?

deepak1725:
E2E: https://ci.centos.org/job/devtools-fabric8-analytics-server-f8a-build-master/377/

PR #656, opened 2020-08-18
feat: changes for user management

prashbnair:


bayesian/utility/user_utils.py
```diff
@@ -22,42 +26,59 @@ def get_user(uuid_val):
...
     except SQLAlchemyError as e:
-        logger.exception(f"Error fetching user with id {uuid_val}")
+        logger.exception("Error fetching user with id %s", uuid_val)
         raise UserException("Error fetching user") from e
 
 
 @retry(reraise=True, stop=tenacity.stop_after_attempt(3), wait=tenacity.wait_fixed(1))
-def create_user(uuid_val, user_source):
-    """Create User."""
+def create_or_update_user(user_id, snyk_api_token, user_source):
```
deepak1725, line 34: Can we move this to `db_gateway.py`

bayesian/api/user_api.py
```diff
@@ -27,22 +27,32 @@ def get_user(user_id):
...
 
-@user_api.route('/<user_id>', methods=['PUT'])
+@user_api.route('', methods=['PUT'])
 @login_required
-def update_user(user_id):
-    """Endpoint for updating user details."""
+def create_or_update_user():
+    """Endpoint for creating or updating user details."""
+    content = request.json
+    user_id = content['user_id']
```
deepak1725, line 41: I think `user_id` will throw `KeyError` if user_id is not in request. Can you please check.

deepak1725, line 41:
```suggestion
    user_id = content.get('user_id')
```

bayesian/api/user_api.py
```diff
@@ -27,22 +27,32 @@ def get_user(user_id):
...
+
     if not user_id:
         raise HTTPError(400, "user id should be present")
 
-    content = request.json
-    user_utils.update_user(user_id, content['snyk_api_token'])
+    snyk_api_token = content['snyk_api_token']
```
deepak1725, line 46: This may again Throw `KeyError` if `snyk_api_token` not in request.

```suggestion
    snyk_api_token = content.get('snyk_api_token')
```

bayesian/api/user_api.py
```diff
@@ -27,22 +27,32 @@ def get_user(user_id):
...
+    if not snyk_api_token:
+        raise HTTPError(400, 'snyk api token should be present')
+
+    if user_utils.is_snyk_token_valid(snyk_api_token):
+        encrypted_api_token = user_utils.encrypt_api_token(snyk_api_token)
+        user_utils.create_or_update_user(user_id, encrypted_api_token.decode(), "SNYK")
+    else:
+        raise HTTPError(400, "Invalid API Token")
+
```
deepak1725, line 55: I suggest to do it in this way. Looks cleaner
```suggestion
if not user_utils.is_snyk_token_valid(snyk_api_token):
     raise HTTPError(400, "Invalid API Token")
 encrypted_api_token = user_utils.encrypt_api_token(snyk_api_token)
 user_utils.create_or_update_user(user_id, encrypted_api_token.decode(), "SNYK")
```

openshift/template.yaml
```diff
@@ -92,6 +97,11 @@ objects:
               secretKeyRef:
                 name: aws-sqs
                 key: aws_secret_access_key
+          - name: ENCRYPTION_KEY_FOR_SNYK_TOKEN
+            valueFrom:
+              secretKeyRef:
+                name: snyk-encryption
+                key: encryption_key_for_snyk_token
```
deepak1725, line 104: Short, Precise, Easy to read
```suggestion
                key: snyk_encryption_token
```

bayesian/utility/user_utils.py
```diff
@@ -22,42 +26,59 @@ def get_user(uuid_val):
...
+            return False
+    except Exception as e:
+        logger.exception("Encountered exception calling Snyk")
+        raise e
+
+
+def encrypt_api_token(snyk_api_token):
+    """Encryption of Api Token."""
+    cipher = Fernet(ENCRYPTION_KEY_FOR_SNYK_TOKEN.encode())
+    return cipher.encrypt(snyk_api_token.encode())
```
arajkumar, line 75: instead of using single key to encrypt all Snyk API Keys, I suggest to derive a key from UUID for each API key insertion. Something like below,

```
store(uuid, snyk_api_key):
  uuid_hash = sha256(uuid)
  snyk_api_key_encrypted = AES(data=snyk_api_key, key=uuid + salt)
  snyk_api_key_hash = sha256(snyk_api_key)

read(uuid):
  snyk_api_key = AES_decrypt(data=snyk_api_key_encrypted, key=uuid + salt)
```

This approach introduces the following columns to the user database, 
```
uuid_hash - SHA256 hashed value of original UUID
synk_api_key_hash - SHA256 hashed value of original Synk API key to check existence of key without decryption
snyk_api_key_encrypted - AES encrypted Synk API key based original UUID and salt which can be used to query snyk API
```
Also the encryption key for each Snyk API token is different and based on UUID., this ensures that compromise of single encryption key will not expose the rest of the user data as well.

bayesian/utility/user_utils.py
```diff
@@ -22,42 +26,59 @@ def get_user(uuid_val):
...
+def create_or_update_user(user_id, snyk_api_token, user_source):
+    """Create or Update User."""
     try:
-        insert_user_stmt = insert(UserDetails).values(user_id=uuid_val, user_source=user_source,
-                                                      created_date=datetime.datetime.now())
-
-        rdb.session.execute(insert_user_stmt)
-        rdb.session.commit()
-        logger.info(f"User created with id {uuid_val}")
+        if snyk_api_token:
```
arajkumar, line 37: Isn't it redundant validation? Request has been already validated by the route handler right?

bayesian/utility/user_utils.py
```diff
@@ -22,42 +26,59 @@ def get_user(uuid_val):
...
-    except SQLAlchemyError as e:
-        rdb.session.rollback()
-        logger.exception(f"Error updating user with id {user_id}")
-        raise UserException("Error updating user") from e
+        response = requests.post(SNYK_API_TOKEN_VALIDATION_URL, json={'api': snyk_api_token},
+                                 headers={'Content-Type': 'application/json'})
+        if response.status_code == 200:
+            return True
+        else:
+            return False
```
arajkumar, line 66:
```suggestion
        return response.status_code == 200
```

bayesian/utility/user_utils.py
```diff
@@ -22,42 +26,59 @@ def get_user(uuid_val):
...
-
-        rdb.session.execute(update_stmt)
-        rdb.session.commit()
-        logger.info(f"User updated with id {user_id}")
-    except SQLAlchemyError as e:
-        rdb.session.rollback()
-        logger.exception(f"Error updating user with id {user_id}")
-        raise UserException("Error updating user") from e
+        response = requests.post(SNYK_API_TOKEN_VALIDATION_URL, json={'api': snyk_api_token},
+                                 headers={'Content-Type': 'application/json'})
```
arajkumar, line 62: `Content-Type` is automatically implied to `application/json` if json named argument is given to requests.post.

bayesian/utility/user_utils.py
```diff
@@ -4,12 +4,16 @@
...
+from cryptography.fernet import Fernet
 
 from sqlalchemy.exc import SQLAlchemyError
-from sqlalchemy import insert, update
+from sqlalchemy.dialects.postgresql import insert
 
 from f8a_worker.models import (UserDetails)
 from bayesian import rdb
+from bayesian.default_config import SNYK_API_TOKEN_VALIDATION_URL
+from bayesian.default_config import ENCRYPTION_KEY_FOR_SNYK_TOKEN
```
arajkumar, line 16:
```suggestion
from bayesian.default_config import SNYK_API_TOKEN_VALIDATION_URL, ENCRYPTION_KEY_FOR_SNYK_TOKEN
```

bayesian/utility/user_utils.py
```diff
@@ -22,42 +26,59 @@ def get_user(uuid_val):
...
-        rdb.session.execute(update_stmt)
-        rdb.session.commit()
-        logger.info(f"User updated with id {user_id}")
-    except SQLAlchemyError as e:
-        rdb.session.rollback()
-        logger.exception(f"Error updating user with id {user_id}")
-        raise UserException("Error updating user") from e
+        response = requests.post(SNYK_API_TOKEN_VALIDATION_URL, json={'api': snyk_api_token},
+                                 headers={'Content-Type': 'application/json'})
+        if response.status_code == 200:
```
arajkumar, line 63: Usual pattern is to use `response.raise_for_status()`

bayesian/utility/user_utils.py
```diff
@@ -22,42 +26,59 @@ def get_user(uuid_val):
...
+            return True
+        else:
+            return False
+    except Exception as e:
+        logger.exception("Encountered exception calling Snyk")
+        raise e
```
arajkumar, line 69: How about using Exception for flow control? it is quite common in Python world.
The given token is valid if there no Exeption, otherwise invalid.

```suggestion
response = requests.post(SNYK_API_TOKEN_VALIDATION_URL, json={'api': snyk_api_token})
response.raise_for_status()
```

bayesian/api/user_api.py
prashbnair: Done

bayesian/api/user_api.py
prashbnair: Done

bayesian/api/user_api.py
prashbnair: Done

bayesian/utility/user_utils.py
prashbnair: Done

bayesian/utility/user_utils.py
prashbnair: I am not sure if we want all DB interactions to go through that. That file would then being to become very big.

bayesian/utility/user_utils.py
prashbnair: Done

bayesian/utility/user_utils.py
prashbnair: Done

bayesian/utility/user_utils.py
deepak1725: I leave it up to you. :)

bayesian/utility/user_utils.py
prashbnair: Can you let me know why we need to store the hash values. Hashing is one way and  the value cannot be retrieved. There is also no comparison to be done against this value so why should it be stored.
If you are using a salt dont you need to remember the salt for decryption.

openshift/template.yaml
prashbnair: It is an encryption key for the API token.

bayesian/utility/user_utils.py
prashbnair: Done

bayesian/utility/user_utils.py
arajkumar: > Can you let me know why we need to store the hash values

Hashing the Synk API key is meant for comparison, if you want to check whether the Snyk API key is already exists during a call to create_or_update user. Also this can be used to report how many users are using same Snyk API key.

> If you are using a salt dont you need to remember the salt for decryption

Yes, salt should be a global configuration(as an env variable). So compromising either salt or uuid can't expose Snyk API key from DB dump.

arajkumar (changes requested):

bayesian/utility/user_utils.py
prashbnair: I am not convinced that this a use case for hashing. We already store the encrypted form and there is no need to do both. For reporting we can always decrypt and figure out how may users are using the same key.

My feeling is that once the salt is leaked it is only a matter of time before it is decrypted. Salts are generally used to ensure that there is no collision when using hashing.

openshift/template.yaml
deepak1725: ok, Can we summarize it in 3 letter variable name..?

bayesian/utility/user_utils.py
arajkumar: Leave about Hashing, what is the problem in using UUID as an encryption key instead of global encryption key?

prashbnair: It just seems like the most obvious key that everybody will use if they want to decrypt.

bayesian/api/user_api.py
prashbnair: Done

bayesian/utility/user_utils.py
arajkumar: Though UUID will be used as key it won't be stored as a plain text in DB., UUID should be hashed and stored.

```
store(uuid, snyk_api_key):
  uuid_hash = sha256(uuid)
  snyk_api_key_encrypted = AES(data=snyk_api_key, key=uuid)

read(uuid):
  snyk_api_key = AES_decrypt(data=snyk_api_key_encrypted, key=uuid)
```

So even if the DB is compromised, data can't be decrypted unless you know each and every UUID.

prashbnair: Whats the problem with storing the key as a secret. Thats exactly what it is for. If the secrets get compromised, then we nothing we add there is safe.

bayesian/utility/user_utils.py
prashbnair: I am not sure I understand this, response.raise_for_status is fine for cases where the request is made. What about the case when the request failed, I would like to print out a log saying the request has failed.

bayesian/utility/user_utils.py
arajkumar: I'm not talking about the key compromising scenario. If database compromised, having a single key for all data would expose the complete data. However when using UUID, each API key will be encrypted using unique key and it broadly reduces the attack surface.

openshift/template.yaml
prashbnair: Is this a standard that we need to follow?

bayesian/utility/user_utils.py
prashbnair: If the key cannot be compromised how can it be decrypted?  I am not seeing any additional value that this can provide.

openshift/template.yaml
deepak1725: 3 letter looks more readable..! If we promote readiabiility, then it can be good standard to follow..

bayesian/utility/user_utils.py
arajkumar: Encrypted data can be compromised using brute force attack. 
With the existing approach, once the attacker gains the key through brute force, then the entire table will be compromised. However in the proposed solution, brute force should be applied for individual rows and would take time.

prashbnair: I understand all of this, for this solution I am not convinced that we need to go through all of this. What we have is good enough for the problem that we have.

openshift/template.yaml
prashbnair: Isn't readability about clearly expressing what the variable means?  The only name that I can think of is encryption_key_snyk_token.

bayesian/default_config.py
```diff
@@ -45,6 +45,11 @@
 BAYESIAN_COMPONENT_TAGGED_COUNT = os.getenv('BAYESIAN_COMPONENT_TAGGED_COUNT', 2)
 
 F8_API_BACKBONE_HOST = os.getenv('F8_API_BACKBONE_HOST', 'http://f8a-server-backbone')
+
+SNYK_API_TOKEN_VALIDATION_URL = os.getenv('SNYK_API_TOKEN_VALIDATION_URL',
+                                          'https://snyk.io/api/v1/verify/token')
+ENCRYPTION_KEY_FOR_SNYK_TOKEN = os.getenv('ENCRYPTION_KEY_FOR_SNYK_TOKEN', 'SNYK')
```
yzainee, line 51: I think you can add this to the snyk secrets

bayesian/utility/user_utils.py
arajkumar: There is a value in data and that is the prime reason for encrypting it. When we choose to encrypt it, why not a hardened approach?. I'm not convinced in going ahead with the existing approach. Thanks.

yzainee: @prashbnair i dont think we need so much of a sophistication for storing the snyk api key. A simple encryption mechanism should be good enough

yzainee (changes requested):

openshift/template.yaml
deepak1725: Fine. Settling with `encryption_key_for_snyk_token` then

bayesian/api/user_api.py
```diff
@@ -22,27 +22,37 @@ def get_user(user_id):
         raise HTTPError(400, "user id should be present")
 
     user = user_utils.get_user(user_id)
-    return jsonify(user_id=user.user_id, status=user.status)
+    user_status = user.status if user.status else UserStatus.FREETIER.name
+    return jsonify(user_id=user.user_id, status=user_status)
 
 
 @user_api.route('', methods=['POST'])
```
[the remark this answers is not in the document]
prashbnair: It is basically what the meaning that POST conveys in REST  which is create. I know we are not creating a user but we are generating a ID. GET ideally should not be doing any of this.

bayesian/default_config.py
prashbnair: Ok

deepak1725 (changes requested):
@prashbnair I have 2 Open Points only if you can let know your take on this, it will be helpful

1. `user_utils.py` don't match our v2 Standards, Why don't we considering shifting v2 code in new standardized named file..?

V2 as a folder exists because we have v2 api's . For the user api, as of now it is not versioned so it stands on its own.

2. I see a Query in `user_utils.py` in `create_or_update_user`.. Why we are not willing to shift it `db_gateway.py`

User Management is a small module with not too much functionality. In case of the API's there is a lot more functionality and it made sense to have a separate data access layer. If I need to move to this to db_gateway it will just be another indirection which is really not required if you look at the code complexity.

deepak1725 (approved):

openshift/template.yaml
```diff
@@ -174,7 +184,7 @@ parameters:
   displayName: Docker image
   required: true
   name: DOCKER_IMAGE
-  value: "openshiftio/bayesian-bayesian-api"
+  value: "prashbnair/bayesian-api"
```
arajkumar, line 187: This should be reverted.

openshift/template.yaml
```diff
@@ -101,7 +111,7 @@ objects:
             value: ${STACK_ANALYSIS_REQUEST_TIMEOUT}
           - name: FLASK_LOGGING_LEVEL
             value: ${FLASK_LOGGING_LEVEL}
-          image: "${DOCKER_REGISTRY}/${DOCKER_IMAGE}:${IMAGE_TAG}"
+          image: "${DOCKER_IMAGE}:${IMAGE_TAG}"
```
arajkumar, line 114: This should be reverted.

bayesian/utility/user_utils.py
arajkumar: @yzainee Whatever I'm proposing is as simple as the one which is used by @prashbnair . The only difference is selection of single key vs value specific key(UUID).

arajkumar (approved):
I'm fine with the changes except the encryption part.

openshift/template.yaml
prashbnair: My mistake, will change

yzainee (approved):
You will need to update the vault secrets before you merge this PR

PR #657, opened 2020-08-18
[APPAI-1447] Part 1 :: User validation based on UUID with decorator

dgpatelgit:
Handle UUID in request header for registered snyk user making stack analyses request.

Jira issue :: https://issues.redhat.com/browse/APPAI-1447

### DIFF
--- a/bayesian/auth.py
+++ b/bayesian/auth.py
@@ -48,9 +48,8 @@ def validate_user(view):
         # By default set this to 'freetier'.
         g.user_status = UserStatus.FREETIER
 
-        # Read uuid from header and try to get user status from RDS.
         uuid = request.headers.get('uuid', None)
-        if uuid is not None:
+        if uuid:
             try:
                 user = get_user(uuid)
             except UserException as e:
--- a/bayesian/auth.py
+++ b/bayesian/auth.py
@@ -46,23 +46,19 @@ def validate_user(view):
         #  ==============================================================
 
         # By default set this to 'freetier'.
-        g.registration_status = UserStatus.FREETIER
+        g.user_status = UserStatus.FREETIER
 
-        # Read uuid from request header.
+        # Read uuid from header and try to get user status from RDS.
         uuid = request.headers.get('uuid', None)
-
-        # Check user with uuid if it is present in RDS.
         if uuid is not None:
             try:
-                # Read user details from RDS based on uuid
                 user = get_user(uuid)
             except UserException as e:
                 logger.warning("Unable to get status for uuid=%s, err=%s", uuid, e)
             else:
-                if user.status == 'REGISTERED':
-                    g.user_status = UserStatus.REGISTERED
+                g.user_status = UserStatus[user.status]
 
-        logger.debug('For UUID: %s, got user type: %s', uuid, g.registration_status)
+        logger.debug('For UUID: %s, got user type: %s', uuid, g.user_status)
         return view(*args, **kwargs)
 
     return wrapper
--- a/bayesian/auth.py
+++ b/bayesian/auth.py
@@ -1,7 +1,10 @@
 """Authorization token handling."""
 import logging
-from flask import request
+from functools import wraps
+from flask import g, request
 from requests import get
+from bayesian.utility.user_utils import get_user
+from bayesian.utility.v2.sa_models import RegistrationStatus
 
 from .default_config import AUTH_URL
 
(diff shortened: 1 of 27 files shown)


bayesian/auth.py
```diff
@@ -26,3 +29,40 @@ def get_access_token(service_name):
 
     except Exception:
         logger.error('Unable to connect to Auth service')
+
+
+def get_user_type(view):
```
arajkumar, line 34: I guess this should be named as `validate_user`

bayesian/auth.py
```diff
@@ -26,3 +29,40 @@ def get_access_token(service_name):
...
+        if uuid is not None:
+            try:
+                # Read user details from RDS based on uuid
+                user = get_user(uuid)
+                if user.status == 'REGISTERED':
+                    g.registration_status = RegistrationStatus.registered
+            except Exception as e:
+                logger.warning('User get for UUID: %s failed with %s', uuid, e)
+
+        logger.debug('For UUID: %s, got user type: %s', uuid, g.registration_status)
```
arajkumar, line 65: Try block should have a block which would really generate an Exception.

```suggestion
if uuid:
    try:
        user = get_user(uuid)
    except <ProperExceptionClass_NOT_GenericException> e:
        logger.warning("Unable to get status for uuid=%s, err=%s", uuid, e)
    else:
        g.registration_status = RegistrationStatus[user.status]
     logger.debug('For UUID: %s, got user type: %s', uuid, g.registration_status)
```

bayesian/utility/v2/sa_models.py
```diff
@@ -31,6 +31,13 @@ class Ecosystem(str, Enum):
     npm = 'npm'
 
 
+class RegistrationStatus(str, Enum):
```
prashbnair, line 34: Can you use the ENUM in user_utils

prashbnair (changes requested):

bayesian/utility/v2/sa_models.py
dgpatelgit: There is difference between status value used in Backbone flow and user utils, we need to have mapping somewhere. We can discuss on this in standup and close it.

dgpatelgit:
> There are two possible solution to generate different response based on user registration status
> 
> 1. Backbone generates response according to registration status
> 2. Backbone can always generate registered user response but API server filters it based on registration status
> 
> Current backbone implementation is based on #1, however #2 seems to be the optimal solution because we can easily switch from unregistered response to registered response without re triggering the full stack analysis.

@arajkumar  As discussed in meeting, we will implement approach #2 and filtering of the report shall be done by API server. I will raise a separate PR with this changes, meanwhile, CA changes needs common implementation of decorator function present in this PR, so I have modified this PR with only user validation decorator changes. PTAL.

bayesian/auth.py
```diff
@@ -26,3 +28,41 @@ def get_access_token(service_name):
...
+        # Check user with uuid if it is present in RDS.
+        if uuid is not None:
+            try:
+                # Read user details from RDS based on uuid
+                user = get_user(uuid)
+            except UserException as e:
+                logger.warning("Unable to get status for uuid=%s, err=%s", uuid, e)
+            else:
+                if user.status == 'REGISTERED':
+                    g.user_status = UserStatus.REGISTERED
```
arajkumar, line 63: Why do we have two formats (types) to represent User status? one is string 'REGISTERED/UNREGISTERED' and another is enum. This should be unified.

bayesian/auth.py
```diff
@@ -26,3 +28,41 @@ def get_access_token(service_name):
...
+        # By default set this to 'freetier'.
+        g.registration_status = UserStatus.FREETIER
+
+        # Read uuid from request header.
+        uuid = request.headers.get('uuid', None)
+
+        # Check user with uuid if it is present in RDS.
+        if uuid is not None:
+            try:
+                # Read user details from RDS based on uuid
```
arajkumar, line 57: I don't recommend to comment on each and every operation unless it is not expressible using code. I feel the code is good enough and comments are redundant.

bayesian/auth.py
arajkumar, line 63: When dealing with enum, you can use a simple approach to map to values.

```suggestion
                g.user_status = UserStatus[user.status]
```

arajkumar (changes requested):

PR #658, opened 2020-08-19
feat: check registered/free tier user

deepak1725:
Check If user is Registered or not based on uuid

Jira: https://issues.redhat.com/browse/APPAI-1253 

### DIFF
--- a/bayesian/utility/db_gateway.py
+++ b/bayesian/utility/db_gateway.py
@@ -21,14 +21,18 @@ import json
 import time
 import logging
 from datetime import datetime
+from typing import Optional, Dict
+
 from requests import post
+from sqlalchemy.orm.exc import NoResultFound, MultipleResultsFound
+
 from bayesian import rdb
 from bayesian.utils import fetch_sa_request, retrieve_worker_result
 
 from sqlalchemy.exc import SQLAlchemyError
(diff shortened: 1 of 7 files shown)


bayesian/utility/db_gateway.py
```diff
@@ -99,6 +103,30 @@ def get_batch_ca_data(cls, ecosystem: str, packages) -> dict:
...
+        return user_details.to_dict()
+
+    except (NoResultFound, MultipleResultsFound):
+        logger.info('No Result found.')
+        return None
+
+    except SQLAlchemyError as e:
+        rdb.session.rollback()
+        logger.exception('%s Error updating log, exception %s', uuid, str(e))
+        raise SQLAlchemyError('Error while checking uuid {}'.format(uuid)) from e
```
arajkumar, line 127: I guess @dgpatelgit is also working on similar concept. Can you guys agree on a common approach to get user registration status?

arajkumar (changes requested):

bayesian/utility/db_gateway.py
prashbnair: +1

deepak1725: @arajkumar Okay will do.  
Why it's a **Requested change** instead of comment..?

PR #664, opened 2020-08-24
changed ingestion error behaviour

deepak1725:
Changed Ingestion Error Behaviour, No Error is thrown when Ingestion is disabled via Flag

Jira: https://issues.redhat.com/browse/APPAI-1454

### DIFF
--- a/bayesian/api/api_v2.py
+++ b/bayesian/api/api_v2.py
@@ -245,7 +245,7 @@ class ComponentAnalysesApi(Resource):
 
             if not stack_recommendation:
                 # If No Package is Known, all are unknown.
-                logger.debug(stack_recommendation)
+                logger.debug(unknown_pkgs)
                 raise HTTPError(202, no_package_available_msg)
 
             if os.environ.get("DISABLE_UNKNOWN_PACKAGE_FLOW") != "1":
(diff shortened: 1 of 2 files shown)


prashbnair:
@deepak1725 From the bug, I understand that we are not returning any data even when there are packages that are present. Is that correct?

deepak1725:
> @deepak1725 From the bug, I understand that we are not returning any data even when there are packages for which we have data. Is that correct?

@prashbnair Previously It was like:
- If there are unknown Packages and INGESTION is DISABLED, we raise an error. Like as mentioned in Bug.

So this PR changes this behavior to:
- If there are unknown Packages and then check if INGESTION is enabled only, trigger Unknown Flow.
No Error, like in Previous case is raised now.

bayesian/api/api_v2.py
```diff
@@ -245,15 +242,17 @@ def post():
...
-            if os.environ.get("DISABLE_UNKNOWN_PACKAGE_FLOW") == "1":
-                # Unknown Packages is Present and INGESTION is DISABLED
-                logger.error(ingestion_disabled_msg)
-                raise HTTPError(400, ingestion_disabled_msg)
-            unknown_package_flow(ecosystem, unknown_pkgs)
+
             if not stack_recommendation:
                 # If No Package is Known, all are unknown.
-                logger.debug(no_package_available_msg)
+                logger.debug(stack_recommendation)
```
prashbnair, line 248: will this not print "None"

deepak1725: Ohh Yeah.. missed it. Fixed.

prashbnair (approved):

PR #685, opened 2020-10-01
fix: double pkg output issue resolved

deepak1725:
The issue was Input Package Order is not maintained by Gremlin which causes invalid Packages versions attached to different objects.

Issue: https://issues.redhat.com/browse/APPAI-1526

### DIFF
--- a/bayesian/utility/v2/component_analyses.py
+++ b/bayesian/utility/v2/component_analyses.py
@@ -124,64 +124,48 @@ def ca_validate_input(input_json: Dict, ecosystem: str) -> Tuple[List[Dict], Lis
 def get_known_unknown_pkgs(
         ecosystem: str, graph_response: Dict,
         normalised_input_pkgs: List) -> Tuple[List[Dict], Set[Package]]:
-    """Analyse Known and Unknown Packages."""
+    """Analyse Known and Unknown Packages.
+
+    :param ecosystem: Ecosystem
+    :param graph_response: Graph Response
+    :param normalised_input_pkgs: Normalised Input Packages
+    :return: Stack Recommendations, Unknown Pkgs
+    """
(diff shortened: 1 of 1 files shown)


bayesian/utility/v2/component_analyses.py
```diff
@@ -125,22 +125,64 @@ def get_known_unknown_pkgs(
...
+    "given_version" and feed that to response builder.
+    :param ecosystem: Ecosystem
+    :param graph_response: Graph Response
+    :param normalised_input_pkgs: Input Packages.
+    :return: Stack Recomm. and DB_Known_Packages.
+    """
+    stack_recommendation = []
+    db_known_packages = set()
+    for package in graph_response.get('result', {}).get('data'):
+        for input_pkg in normalised_input_pkgs:
```
prashbnair, line 175: Can we change this to a map and use the same code for all ecosystems? This will also result in this loop being O(n)

prashbnair (changes requested):

bayesian/utility/v2/component_analyses.py
deepak1725: Mapping added for normalised_input_pkgs in the case of Golang.

> use the same code for all ecosystems.

IMO mapping for other ecosystems doesn't really a Gain here. https://github.com/fabric8-analytics/fabric8-analytics-server/blob/82fd64e9c7f68c280cf6f264730a5366a87a09b1/bayesian/utility/v2/component_analyses.py#L150 

PTAL

prashbnair: we can get rid of duplicate code as there is not much difference in terms time.

deepak1725: Ack.

prashbnair (approved):

PR #687, opened 2020-10-05
[APPAI-1525] CA Response:  free-tier (In case of only security advisory) should not have the recommended version in CA response

dgpatelgit:
For freetier user , hide recommended version for '0' public vulnerability.

Jire Ticket :: https://issues.redhat.com/browse/APPAI-1525


deepak1725 (changes requested):

dgpatelgit:
As discussed during release meeting, we don't need this change as of now. Aborting this PR as of now.

PR #689, opened 2020-10-23
fix:fixed defect for expired users

prashbnair:


yzainee:
@prashbnair When does a user expire? Can you give me an example.

yzainee (approved):
Good to merge after the lint error is removed.

PR #700, opened 2020-11-05
feat: [APPAI-1541] [CA]Handle pseudo version parsing and querying as per new gremlin query

dgpatelgit:
Support for pseudo version for golang component analyses request as per below ticket.

Jira Ticket :: https://issues.redhat.com/browse/APPAI-1541

### DIFF
--- a/bayesian/utility/db_gateway.py
+++ b/bayesian/utility/db_gateway.py
@@ -17,7 +17,6 @@
 """All communications with Graph DB and RDS from API v2 are done here."""
 
 import os
-import re
 import json
 import time
 import logging
@@ -158,7 +157,7 @@ class GraphAnalyses:
         for vuln in vulnerabilities:
             package_name = vuln['package_name'][0]
             if gh._is_commit_date_in_vuln_range(
-               GraphAnalyses.extract_timestamp(
+               gh.extract_timestamp(
                    package_version_map[package_name]), vuln['vuln_commit_date_rules'][0]):
                 if package_name not in filter_vulnerabilities:
                     filter_vulnerabilities[package_name] = []
@@ -193,7 +192,7 @@ class GraphAnalyses:
             data.append({
                 'package': pckg,
                 'version': {},
-                'vuln': vulnerabilities.get(pckg['name'][0], [])
+                'cve': vulnerabilities.get(pckg['name'][0], [])
             })
         pckg_response['result']['data'] = data
 
@@ -201,15 +200,6 @@ class GraphAnalyses:
         logger.info("It took %s to fetch pseudo version results.", elapsed_time)
 
         return pckg_response
-
-    @classmethod
-    def extract_timestamp(cls, in_string: str) -> str:
-        """Extract timestamp value YYYYMMDDHHMMSS from given string."""
-        timestamp = re.findall(r'\d{14}', in_string)
-        if len(timestamp) > 0:
-            return timestamp[0]
-
-        return None
 
 
 class RdbAnalyses:
--- a/bayesian/utility/db_gateway.py
+++ b/bayesian/utility/db_gateway.py
@@ -149,7 +149,7 @@ class GraphAnalyses:
         return response.json()
 
     @classmethod
-    def filter_vulnerable_packages(vulnerabilities, package_version_map) -> dict:
+    def filter_vulnerable_packages(cls, vulnerabilities: list, package_version_map: dict) -> dict:
         """Filter vulnerable package based on timestamp in pseudo version."""
         logger.debug('Executing filter_vulnerable_packages')
 
@@ -177,7 +177,7 @@ class GraphAnalyses:
         package_version_map = {}
         for pckg in packages:
             package_name = pckg['name'].split('@')[0]
-            filter_packages.append(package_name)
+            filter_packages.add(package_name)
             package_version_map[package_name] = pckg['version']
 
         vuln_response = GraphAnalyses.get_vulnerabilities_for_packages(
@@ -200,7 +200,7 @@ class GraphAnalyses:
         elapsed_time = time.time() - started_at
         logger.info("It took %s to fetch pseudo version results.", elapsed_time)
 
-        return pckg_response.json()
+        return pckg_response
 
     @classmethod
     def extract_timestamp(cls, in_string: str) -> str:
--- a/bayesian/utility/db_gateway.py
+++ b/bayesian/utility/db_gateway.py
@@ -90,15 +90,11 @@ class GraphAnalyses:
         return query_result.json()
 
     @classmethod
-    def get_batch_ca_data(cls, ecosystem: str, packages) -> dict:
-        """Component Analyses Batch Call."""
-        logger.debug('Executing get_batch_ca_data')
+    def post_gremlin(cls, query: str, bindings: dict = None) -> dict:
+        """Post the given query and bindings to gremlin endpoint."""
         payload = {
-            'gremlin': cls.ca_batch_query,
-            'bindings': {
-                'ecosystem': ecosystem,
-                'packages': packages
-            }
+            'gremlin': query,
+            'bindings': bindings
         }
         started_at = time.time()
         response = post(url=gremlin_url, data=json.dumps(payload))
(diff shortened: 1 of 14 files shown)


arajkumar (changes requested):

bayesian/utility/db_gateway.py
```diff
@@ -98,6 +108,80 @@ def get_batch_ca_data(cls, ecosystem: str, packages) -> dict:
...
+
+        payload = {
+            'gremlin': cls.ca_batch_vuln_query,
+            'bindings': {
+                'ecosystem': ecosystem,
+                'packages': list(filter_packages)
+            }
+        }
+        started_at = time.time()
+        vuln_response = post(url=gremlin_url, data=json.dumps(payload))
```
yzainee, line 131: use try catch here

bayesian/utility/db_gateway.py
```diff
@@ -98,6 +108,80 @@ def get_batch_ca_data(cls, ecosystem: str, packages) -> dict:
...
+
+        payload = {
+            'gremlin': cls.ca_batch_package_query,
+            'bindings': {
+                'ecosystem': ecosystem,
+                'packages': list(vulnerabilities.keys())
+            }
+        }
+        started_pckg_at = time.time()
+        pckg_response = post(url=gremlin_url, data=json.dumps(payload))
```
yzainee, line 155: try catch needed here

bayesian/utility/v2/component_analyses.py
```diff
@@ -190,3 +225,17 @@ def add_unknown_pkg_info(stack_recommendation: List, unknown_pkgs: Set[Package])
...
+def is_pseudo_version(ecosystem: str, version: str) -> bool:
+    """Check if given version is a pseudo version for golang ecosystem.
+
+    :param ecosystem: str
+    :param version: str - clean package version.
+    :return: True is its pseudo version for golang otherwise False.
+    """
+    if ecosystem == "golang":
+        # It is pseudo if contains 14 numeric timestamp and 12 digits commit hash.
+        return len(re.findall(r'(\d{14}|[0-9a-zA-Z]{12})', version)) == 2
```
yzainee, line 239: between the timestamp and the hash, we have a '-'. You can use that for the complete regex

bayesian/utility/v2/component_analyses.py
```diff
@@ -114,13 +119,50 @@ def ca_validate_input(input_json: Dict, ecosystem: str) -> Tuple[List[Dict], Lis
...
+    logger.debug('Executing get_batch_ca_data')
+
+    aggregated_response = {}
+    semver_packages = []
+    pseudo_version_packages = []
+
+    # Need to seperate semver and pseudo verion packages for golang
+    if (ecosystem == "golang"):
+        for p in packages:
+            if p['is_pseudo_version']:
```
yzainee, line 143: Wont you get only modules from here? packages wont have that pseudo version attached. If yes, then your previous query etc based on the 'package_name' will be incorrect.

yzainee (changes requested):

bayesian/utility/v2/component_analyses.py
dgpatelgit: 'is_pseudo_version' is a new flag set by this code for each package object (which has package and module data). If you check 'ca_validate_input()' above this code, pseudo version flag is set for every package in the request based on version string in the request.

bayesian/utility/db_gateway.py
dgpatelgit: next statement, raise_for_status() takes care of HTTP exception of 4XX and 5XX errors.

bayesian/utility/db_gateway.py
dgpatelgit: next statement, raise_for_status() takes care of HTTP exception of 4XX and 5XX errors.

bayesian/utility/db_gateway.py
```diff
@@ -98,6 +108,80 @@ def get_batch_ca_data(cls, ecosystem: str, packages) -> dict:
         logger.info("It took %s to fetch results from Gremlin.", elapsed_time)
         return response.json()
 
+    @classmethod
+    def get_batch_ca_data_for_pseudo_version(cls, ecosystem: str, packages) -> dict:
+        """Component analyses batch call only for pseudo version applicable for golang."""
```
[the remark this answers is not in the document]
dgpatelgit: Duplicate comment same as previous. Ignoring it.

dgpatelgit (commented):
Implemented first level of comments, few justified with details. PTAL.

bayesian/utility/v2/component_analyses.py
yzainee: @dgpatelgit Thats the point im trying to make. I want to understand the reason for fetching using the package name instead of module name. Whats the reason behind it? pseudo versions are added at the module level.

dgpatelgit: Vulnerability node has "module_name" and "package_name", so instead of getting all vulns for module and then filtering response using package_name, I thought of getting vuln using "package_name" itself. As you mentioned in the slack during discussion, that for vuln at module level, vuln node will have module_name == pacakge_name.

yzainee: ok.I get what you are trying to do here. Instead of module you want to query package. That will work. no problems. Please cross verify the response time. As the number of packages will be more, it might take more time to get response from gremlin. So please gauge both the approaches and select which suits best.

bayesian/utility/db_gateway.py
```diff
@@ -56,6 +58,14 @@ class GraphAnalyses:
             .valueMap().fold()).fill(epv);};epv;
             """
 
+    get_vuln_query = """
+            g.V().has('snyk_ecosystem', ecosystem).has('package_name', within(packages)).valueMap()
+            """
+
+    get_package_query = """
+            g.V().has('ecosystem', ecosystem).has('name', within(packages)).valueMap()
```
deepak1725, line 66: Don't we want to query `synk_ecosystem` node..?

yzainee: This is a query for the package node not for vulnerability node. @deepak1725

dgpatelgit: We query `snyk_ecosystem` node only for getting vulnerabilities, this query is to get package details for vulnerable packages.

bayesian/utility/v2/component_analyses.py
```diff
@@ -190,3 +225,17 @@ def add_unknown_pkg_info(stack_recommendation: List, unknown_pkgs: Set[Package])
         unknowns.pop('given_version', None)
         stack_recommendation.append(unknowns)
     return stack_recommendation
+
+
+def is_pseudo_version(ecosystem: str, version: str) -> bool:
```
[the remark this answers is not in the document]
dgpatelgit: This is made generic and moved to Utils.

bayesian/utility/db_gateway.py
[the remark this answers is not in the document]
dgpatelgit: This is broken into many smaller functions.

bayesian/utility/v2/component_analyses.py
dgpatelgit: Regex is fixed now to validate 0.0.0-timestamp-commithash and its moved to UTILs.

bayesian/utility/v2/component_analyses.py
dgpatelgit: I have check  the performance on module_name and package_name queries, on a small sample data I noticed that module_name took around 350ms to get vuln data for modules and all its packages, on top API server need to filter required packages from response. Whereas second query based on package_name took 250ms and completely avoided response filtering in API server. So with this data, I took a call to fetch gremlin vuln data based on package_name filter.

bayesian/utility/db_gateway.py
```diff
@@ -98,6 +107,100 @@ def get_batch_ca_data(cls, ecosystem: str, packages) -> dict:
         logger.info("It took %s to fetch results from Gremlin.", elapsed_time)
         return response.json()
 
+    @classmethod
+    def get_vulnerabilities_for_packages(cls, ecosystem: str, packages) -> dict:
```
arajkumar, line 111: There is a quite a bit of code duplicate between get_vulnerabilities_for_packages and get_package_details. Can you create a generic util function which takes just gremlin query and binding like below?

https://github.com/fabric8-analytics/f8a-server-backbone/blob/master/src/utils.py#L325

bayesian/utility/db_gateway.py
```diff
@@ -98,6 +107,100 @@ def get_batch_ca_data(cls, ecosystem: str, packages) -> dict:
...
+            vuln_response.get('result', {}).get('data', []), package_version_map)
+        pckg_response = GraphAnalyses.get_package_details(
+            ecosystem, list(vulnerabilities.keys()))
+
+        # Merge the package and vunlerability data into response.
+        data = []
+        for pckg in pckg_response.get('result', {}).get('data', []):
+            data.append({
+                'package': pckg,
+                'version': {},
```
arajkumar, line 194: Should the version field populated using package_version_map?, else why it is been kept empty?

arajkumar (changes requested):

bayesian/utility/db_gateway.py
dgpatelgit: Pseudo based packages don't have version node in graph db. Also we don't need this in CA, So to explicitly indicate that Pseudo packages don't have 'version', I have set it to empty dict.

bayesian/utility/db_gateway.py
dgpatelgit: This is done.

bayesian/utility/db_gateway.py
arajkumar: The CA response should have a version right?, even for pseudo version?

dgpatelgit: In response version is filled by 'given_version' that is read from request. We don't read version from Graph as of now.

arajkumar (approved):
lgtm

yzainee (approved):
LGTM

PR #702, opened 2020-11-26
chore: [APPAI-1639] Modified package name to include module name and include the same in response

dgpatelgit:
This PR synchronised the package name format between CA and SA request for golang. Package name includes module like package_name@module_name. CA response should also include package name in this format, this will help plugin to simplify their logic of consuming CA data and forming aggregated diagnostic message.

Jira Ticket :: https://issues.redhat.com/browse/APPAI-1639

### DIFF
--- a/bayesian/utility/v2/component_analyses.py
+++ b/bayesian/utility/v2/component_analyses.py
@@ -33,7 +33,7 @@ from werkzeug.exceptions import BadRequest
 from bayesian.utility.db_gateway import GraphAnalyses
 
 logger = logging.getLogger(__name__)
-Package = namedtuple("Package", ["name", "version", "package_unknown",
+Package = namedtuple("Package", ["name", "given_name", "version", "package_unknown",
                                  "given_version", "is_pseudo_version"])
 
 
@@ -45,12 +45,14 @@ def validate_version(version: str) -> bool:
     return True
 
 
-def normlize_packages(package: str, version: str, given_version: str,
+def normlize_packages(name: str, given_name: str,
+                      version: str, given_version: str,
                       is_pseudo_version: bool) -> Package:
     """Normalise Packages into hashable."""
     logger.debug('Normalizing Packages.')
     return Package(
-        name=package, version=version, given_version=given_version,
+        name=name, given_name=given_name,
+        version=version, given_version=given_version,
         is_pseudo_version=is_pseudo_version, package_unknown=True)
 
 
@@ -99,7 +101,7 @@ def ca_validate_input(input_json: Dict, ecosystem: str) -> Tuple[List[Dict], Lis
     normalised_input_pkgs = []
     for pkg in input_json.get('package_versions'):
         pseudo_version = False
-        package = pkg.get("package")
+        package = given_package = pkg.get("package")
         clean_version = given_version = pkg.get("version")
         if not all([package, given_version]):
             error_msg = "Invalid Input: Package, Version are required."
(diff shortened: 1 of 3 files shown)


bayesian/utility/v2/component_analyses.py
```diff
@@ -122,11 +124,14 @@ def ca_validate_input(input_json: Dict, ecosystem: str) -> Tuple[List[Dict], Lis
         if ecosystem == 'golang':
             _, clean_version = GolangDependencyTreeGenerator.clean_version(given_version)
             pseudo_version = gh.is_pseudo_version(clean_version)
+            # Strip module appended to the package name
+            package = package.split('@')[0]
 
         packages_list.append(
-            {"name": package, "version": clean_version, "given_version": given_version,
+            {"name": package, "given_name": given_package,
+             "version": clean_version, "given_version": given_version,
```
prashbnair, line 132: Does given_name need to be passed into this? It does not seem to be used.

arajkumar (approved):
It is upto you to consider the change proposed in the review comment.

Rest looks ok to me.

prashbnair (changes requested):

deepak1725 (approved):
+1 to Arun's Change. 

Recommendation: Raise a Refactor Issue and take it just after the release.

The stuff looks polluted here.

bayesian/utility/v2/component_analyses.py
```diff
@@ -45,12 +45,14 @@ def validate_version(version: str) -> bool:
     return True
 
 
-def normlize_packages(package: str, version: str, given_version: str,
+def normlize_packages(name: str, given_name: str,
+                      version: str, given_version: str,
                       is_pseudo_version: bool) -> Package:
```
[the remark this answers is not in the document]
dgpatelgit: @arajkumar can you raise a ticket, so that we can have this in our refactoring backlog?

bayesian/utility/v2/component_analyses.py
dgpatelgit: This is used while building response. At line number 190, we are building a map with this value. This map is used by get_given_name_and_version() at line number 230, to extract given_name. The given_name is then passed to generate_recommendation() to actually build the response.

prashbnair: I was referring to packages_list which only seems to be used in get_batch_ca_data. I do not see packages_list being used anywhere else.

dgpatelgit: 'packages_list' is existing flow and is used very much, its returned from this function to the caller (api_v2:227), which again pass this list to get response (api_v2:229).

prashbnair: Does this function use any of the parameters that you are passing
https://github.com/fabric8-analytics/fabric8-analytics-server/blob/master/bayesian/api/api_v2.py#L229

Does this function use package_list
https://github.com/fabric8-analytics/fabric8-analytics-server/blob/master/bayesian/api/api_v2.py#L231

dgpatelgit: Yes, both of this function use all parameters passed to it. Do you see otherwise?

prashbnair (approved):

bayesian/utility/v2/component_analyses.py
prashbnair: Please go ahead and merge, we can have this discussion offline. The code needs to refactored a lot.

PR #704, opened 2020-12-01
Minor fixes to the user API

prashbnair:
This PR fixes
- Invalid Synk Token HTML in response error
- Invalid Default value for `ENCRYPTION_KEY_FOR_SNYK_TOKEN` in config files
- Empty Payload: Server error in user [PUT] API.

### DIFF
--- a/bayesian/api/user_api.py
+++ b/bayesian/api/user_api.py
@@ -10,7 +10,7 @@ from f8a_utils.user_token_utils import is_snyk_token_valid, encrypt_api_token, U
 
 from bayesian.exceptions import HTTPError
 from bayesian.utility import user_utils
-from bayesian.utility.user_utils import UserException, UserNotFoundException
+from bayesian.utility.user_utils import UserException
 
 user_api = Blueprint('user_api', __name__, url_prefix='/user')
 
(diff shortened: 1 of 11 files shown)


bayesian/default_config.py
```diff
@@ -48,7 +48,8 @@
 
 SNYK_API_TOKEN_VALIDATION_URL = os.getenv('SNYK_API_TOKEN_VALIDATION_URL',
                                           'https://snyk.io/api/v1/verify/token')
-ENCRYPTION_KEY_FOR_SNYK_TOKEN = os.getenv('ENCRYPTION_KEY_FOR_SNYK_TOKEN', 'SNYK')
+ENCRYPTION_KEY_FOR_SNYK_TOKEN = os.getenv('ENCRYPTION_KEY_FOR_SNYK_TOKEN',
```
deepak1725, line 51: `user[PUT]` API is not using this value here but one in https://github.com/fabric8-analytics/fabric8-analytics-utils/blob/e2207358ecb8289e2e941b0011d8b86c848e8a21/f8a_utils/default_config.py#L6

deepak1725: Although it's good to have this change here too.

bayesian/api/user_api.py
```diff
@@ -40,17 +42,20 @@ def generate_uuid_for_user():
 def create_or_update_user():
     """Endpoint for creating or updating user details."""
     content = request.json
```
deepak1725, line 44: can we check for empty payload here..?

deepak1725 (changes requested):

bayesian/default_config.py
prashbnair: Done

bayesian/api/user_api.py
prashbnair: It is being checked.  The content is null condition takes care of it.

deepak1725: Sorry, I missed it.

deepak1725 (approved):
lgtm

PR #705, opened 2020-12-07
fix: [CA] Trigger unknown package using Ingestion API

jparsai:
# Description 
Using Ingestion API to trigger worker flow instead of embedded code, also enabled Golang ingestion jobs.

# Type of changes
- [x] Used function and data type from Utils for unknown package ingestion. 
- [x] Updated Utils and Worker versions.
- [x] Removed Env variables and check for unknown flow ingestion as it is handled in Ingestion API.  
- [x] Modified some test cases.
- [x] Removed some test cases used for checking execution based in value of env variables.


openshift/template.yaml
```diff
@@ -111,6 +111,11 @@ objects:
             value: ${STACK_ANALYSIS_REQUEST_TIMEOUT}
           - name: FLASK_LOGGING_LEVEL
             value: ${FLASK_LOGGING_LEVEL}
+          - name: APP_SECRET_KEY
+            valueFrom:
+              secretKeyRef:
+                name: jobs
+                key: flask-app-secret-key
```
arajkumar, line 118: This looks so generic? BTW, why do you need a secrete to access internal service?

bayesian/utility/v2/component_analyses.py
```diff
@@ -58,12 +67,28 @@ def normlize_packages(name: str, given_name: str,
 
 def unknown_package_flow(ecosystem: str, unknown_pkgs: Set[namedtuple]) -> bool:
```
arajkumar, line 68: Though it is quite small, why can't this be considered as a util function? If it is, shall be shared with backbone as well.

bayesian/api/api_v2.py
```diff
@@ -184,10 +183,16 @@ def get(ecosystem, package, version):
 
             return response_template({'error': msg}, 202)
 
+        # Creating a set to be passed to the function unknown_package_flow
+        unknown_pkgs = set()
+        unknown_pkgs.add(normlize_packages(name=package, given_name=package,
```
arajkumar, line 188: not necessary to use normalized_packages. There is a package model which you can use which has only name and version

arajkumar (changes requested):

openshift/template.yaml
jparsai: In initial discussion about Ingestion API it was decided that we will not use any token which can will expire after some time, so we can use a secret for validation.
Yes it is used internally as of now so validation is not necessary, but it is part of Jobs APIs and its route is exposed outside, so it would be to keep some kind of authentication.

bayesian/utility/v2/component_analyses.py
jparsai: Yes, I am Okay to keep it in Utils, if team is fine with it, I can move it.

bayesian/api/api_v2.py
jparsai: Actually I did it this way because in POST API this format is been used and then object has been used further, I wanted to use same function in GET & POST APIs as currently it has separate function to trigger flow for GET and POST.

openshift/template.yaml
arajkumar: I would suggest two endpoints, one for internal and another for external(both can call same function). From the Route description you can use `host` property for external api. Just a thought.

@yzainee

bayesian/utility/v2/component_analyses.py
jparsai: @yzainee Please give your comment on it.

openshift/template.yaml
yzainee: what advantage would that have? having 2 APIs doing the same stuff?

bayesian/utility/v2/component_analyses.py
yzainee: As its not directly interacting with gremlin and all, i am ok with the idea of moving it to utils. please make sure you add a separate file for it with added test cases. utils has 95% test cov criteria

openshift/template.yaml
arajkumar: @yzainee No shared secrete hassle involved

yzainee: I am not for or against having the 2 endpoints. Will leave it upto you how you want to go ahead.

jparsai: Yes, I will do that as it was discussed in meetings and everyone was in agreement. I have not done that yet because this is already being used by repos like reports and github refresh, If I change API auth then I need to update it in other repos immediately, that is why I have not done any changes in endpoints as of now, I will create new end points and one by one I will update them in all consumer repos. so it wont block us from starting unknown flow ingestion and github refresh, unknwon version ingestion will also work fine, in the meantime we introduce new end point without authentication.

jparsai: @yzainee @arajkumar if you are Okay with it and if there are no code changes required, can you please approve it?

yzainee (approved):

bayesian/api/api_v2.py
```diff
@@ -184,10 +184,16 @@ def get(ecosystem, package, version):
...
```
arajkumar, line 189: You may have to use correct type here. ingestion_utils.Packages?

bayesian/api/api_v2.py
```diff
@@ -241,7 +247,7 @@ def post():
         # Step4: Handle Unknown Packages
         if unknown_pkgs:
             stack_recommendation = add_unknown_pkg_info(stack_recommendation, unknown_pkgs)
-            if os.environ.get("DISABLE_UNKNOWN_PACKAGE_FLOW") != "1" and ecosystem != "golang":
+            if os.environ.get("DISABLE_UNKNOWN_PACKAGE_FLOW") != "1":
```
arajkumar, line 250: Can you do the same here as backbone? i.e. letting the jobs to decide whether to ingest based on environment(staging?)

arajkumar (changes requested):

bayesian/api/api_v2.py
```diff
@@ -247,10 +219,11 @@ def post():
...
         if unknown_pkgs:
             stack_recommendation = add_unknown_pkg_info(stack_recommendation, unknown_pkgs)
-            if os.environ.get("DISABLE_UNKNOWN_PACKAGE_FLOW") != "1":
-                # Unknown Packages is Present and INGESTION is Enabled
-                logger.debug(unknown_pkgs)
-                unknown_package_flow(ecosystem, unknown_pkgs)
+            pkgs_to_ingest = set(map(lambda pkg: ingestion_utils.Package(package=pkg.package,
+                                                                         version=pkg.version),
+                                     unknown_pkgs))
+            logger.debug(pkgs_to_ingest)
```
arajkumar, line 225: Add some info on what you are logging.
```suggestion
            logger.debug("Unknown ingestion triggered for %s", pkgs_to_ingest)
```

arajkumar (changes requested):

arajkumar (approved):
lgtm

PR #708, opened 2020-12-09
chore: Remove reference to requestJson column from SA requests table

rafiu007:
The requestJson column has been removed from SA table as it contains duplicate data.[https://github.com/fabric8-analytics/fabric8-analytics-worker/pull/926](url).


rafiu007:
No need for this pr any more as we are not removing the column requestJson.

PR #709, opened 2020-12-11
fix: [APPAI-1659] CA gives Unknown Package when there are same packages with different version

dgpatelgit:
Jira ticket :: https://issues.redhat.com/browse/APPAI-1659


arajkumar (approved):
lgtm

PR #713, opened 2021-01-04
fix: Persist uuid from SA call in `stack_analyses_request` table

rafiu007:
This pr makes chnages in the api server so it can store the uuid from sa calls into the newly created user_id column in stack_analyses_request table.

### DIFF
--- a/bayesian/utility/db_gateway.py
+++ b/bayesian/utility/db_gateway.py
@@ -31,6 +31,7 @@ from sqlalchemy.dialects.postgresql import insert
 
 from f8a_worker.models import StackAnalysisRequest
 
+
 logger = logging.getLogger(__name__)
 gremlin_url = "http://{host}:{port}".format(
     host=os.environ.get("BAYESIAN_GREMLIN_HTTP_SERVICE_HOST", "localhost"),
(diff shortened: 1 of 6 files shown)


bayesian/utility/v2/stack_analyses.py
```diff
@@ -55,13 +57,15 @@ def post_request(self):
         # Generate unique request id using UUID, also record timestamp in readable form
         self._new_request_id = str(uuid.uuid4().hex)
         date_str = str(datetime.datetime.now())
+        # Fetch uuid from header
+        header_data = HeaderData(uuid=request.headers.get('uuid', None))
```
prashbnair, line 61: why do you need to use HeaderData. you can store the uuid in a variable.

bayesian/utility/v2/stack_analyses.py
```diff
@@ -55,13 +57,15 @@ def post_request(self):
...
 
         # Make backbone request
         deps = self._make_backbone_request()
 
         # Finally save results in RDS and upon success return request id.
         rdbAnalyses = RdbAnalyses(self._new_request_id, date_str,
-                                  self._manifest_file_info, deps)
+                                  self._manifest_file_info, deps, header_data.uuid)
```
prashbnair, line 68: where is the code to save the field.

bayesian/utility/db_gateway.py
```diff
@@ -217,12 +217,13 @@ class RdbAnalyses:
     Provides interfaces to save and read request post data for stack analyses v2.
     """
 
-    def __init__(self, request_id, submit_time=None, manifest=None, deps=None):
+    def __init__(self, request_id, submit_time=None, manifest=None, deps=None, user_id=None):
```
prashbnair, line 220:
```suggestion
    def __init__(self, request_id):
```

Change the constructor to only take in parameters that are used by all functions.  The other parameters are specific to the save_post_request method and can be passed as parameters to that.

prashbnair (changes requested):

bayesian/utility/db_gateway.py
rafiu007: Okay .Makes sense.

rafiu007: Resolved.

prashbnair (approved):

PR #715, opened 2021-01-11
fix: Add book keeping for CA post call to persist uuid in `api_requests1 table.

rafiu007:


bayesian/utils.py
```diff
@@ -103,6 +103,21 @@ def server_create_component_bookkeeping(ecosystem, name, version, user_profile):
     return server_run_flow('componentApiFlow', args)
 
 
+def server_create_component_bookkeeping_post(ecosystem, header):
```
prashbnair, line 106: Can this be called create_component_analysis_bookkeeping. I dont think the post is required. Like I said in the worker PR if the user_email and user_profile are always 'None' there is no point sending it.

prashbnair (changes requested):

bayesian/utils.py
rafiu007: Yes,I will change accordingly.

rafiu007:
The team has decided to create a new table for storing data from CA calls.

PR #720, opened 2021-02-02
fix:Add bookkeeping task for CA post calls.

rafiu007:
This bookkeeping task will store data into RDS from CA post calls.

### DIFF
--- a/bayesian/api/api_v2.py
+++ b/bayesian/api/api_v2.py
@@ -39,6 +39,7 @@ from bayesian.utility.v2.component_analyses import ca_validate_input, \
     get_known_unknown_pkgs, add_unknown_pkg_info, get_batch_ca_data
 from bayesian.utils import (get_system_version,
                             server_create_component_bookkeeping,
+                            create_component_bookkeeping,
                             check_for_accepted_ecosystem)
 from bayesian.utility.v2.ca_response_builder import ComponentAnalyses
 from bayesian.utility.v2.sa_response_builder import (StackAnalysesResponseBuilder,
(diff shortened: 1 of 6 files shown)


bayesian/utils.py
```diff
@@ -103,6 +103,20 @@ def server_create_component_bookkeeping(ecosystem, name, version, user_profile):
     return server_run_flow('componentApiFlow', args)
 
 
+def create_component_bookkeeping(ecosystem, packages_list, user_id, user_agent):
+    """Run the component analysis for given ecosystem+package+version."""
+    args = {
+        'api_name': 'component_analyses_post',
+        'external_request_id': uuid.uuid4().hex,
```
prashbnair, line 110: My suggestion will be to generate this on the client side and use this for all the POST requests for a single stack. With this we will be able to figure out the stack.

Are we not populating the manifest_hash

bayesian/api/api_v2.py
```diff
@@ -39,6 +39,7 @@
     get_known_unknown_pkgs, add_unknown_pkg_info, get_batch_ca_data
 from bayesian.utils import (get_system_version,
                             server_create_component_bookkeeping,
```
prashbnair, line 41: Is this component being used?

prashbnair (changes requested):

bayesian/api/api_v2.py
rafiu007: when CA get call is made that is used to store data in api_requests table.

prashbnair: I dont think we need to track that anymore.

bayesian/utils.py
rafiu007: For the manifest_hash I will add codeto fetch it from the user.

bayesian/api/api_v2.py
rafiu007: I will modify accordingly.

PR #727, opened 2021-02-09
fix: Add function to store data from CA post calls.

rafiu007:


bayesian/utils.py
```diff
@@ -64,16 +63,19 @@ def get_user_email(user_profile):
         return default_email
 
 
-def server_create_component_bookkeeping(ecosystem, name, version, user_profile):
+def create_component_bookkeeping(ecosystem, packages_list,
+                                 user_id, user_agent, manifest_hash, request_id):
     """Run the component analysis for given ecosystem+package+version."""
     args = {
-        'external_request_id': uuid.uuid4().hex,
         'data': {
```
prashbnair, line 70: Why are the other fields not in the data dict?

prashbnair (changes requested):

bayesian/utils.py
rafiu007: No particular reason.I ll move it.

prashbnair (approved):

bayesian/api/api_v2.py
```diff
@@ -199,6 +197,9 @@ def post():
         response_template: Tuple = namedtuple("response_template", ["message", "status", "headers"])
         input_json: Dict = request.get_json()
         ecosystem: str = input_json.get('ecosystem')
+        user_agent = request.headers.get('User-Agent', None)
+        manifest_hash = str(request.headers.get('manifest_hash', None))
```
arajkumar, line 201: Why are we overloading request header? shouldn't this be optional filed of request payload?

bayesian/api/api_v2.py
```diff
@@ -199,6 +197,9 @@ def post():
...
+        request_id = request.headers.get('request_id', None)
```
arajkumar, line 202: Why are we overloading request header? shouldn't this be optional filed of request payload?

bayesian/utils.py
```diff
@@ -64,15 +63,18 @@ def get_user_email(user_profile):
...
     args = {
-        'external_request_id': uuid.uuid4().hex,
+        'external_request_id': request_id,
         'data': {
-            'api_name': 'component_analyses',
-            'user_email': get_user_email(user_profile),
-            'user_profile': user_profile,
-            'request': {'ecosystem': ecosystem, 'name': name, 'version': version}
+            'api_name': 'component_analyses_post',
+            'manifest_hash': hashlib.md5(manifest_hash.encode()).hexdigest(),
```
arajkumar, line 73: The field name in the request says manifest_hash, then why are computing again here?

rafiu007: Manifeat_hash will actually contain the file path.

bayesian/api/api_v2.py
rafiu007: @prashbnair

bayesian/utils.py
rafiu007: Resolved.

bayesian/utils.py
rafiu007: when in the header.

bayesian/api/api_v2.py
prashbnair: I dont think any of these fields are optional. I am now having a doubt if we should capture the path as it reveals some data about the user. I read somewhere that it is not allowed.

arajkumar: we can apply the hash at client level. however still SA request has manifest path.

bayesian/utils.py
arajkumar: it should be fixed.

also it needs api spec change

PR #728, opened 2021-02-09
chore: Add default coreapi-release to keep docker build happy

arajkumar:
This fixes a problem when running `docker build`. It is not an issue in CI/CD as it uses makefiles which handles this case., it is just to make developer life happy.


hack/coreapi-release-template
```diff
@@ -0,0 +1,3 @@
+COMMITTED_AT=1970-01-01 00:00:00 +0000
+COMMIT_HASH=badbadf
```
deepak1725, line 2:
```suggestion
COMMIT_HASH=randomhash
```

deepak1725 (changes requested):

hack/coreapi-release-template
arajkumar: Your suggestion is not a valid time stamp and hash.

arajkumar: It is verified from smoke test

deepak1725: Isn't it a placeholder..?

arajkumar: when running smoke test on images build using `docker build` should pass right?
e.g. 
1. you built image using `docker build . -t foo` and deployed on your dev cluster
2. executing smoke test on the dev cluster should pass

deepak1725 (approved):

sawood14012 (approved):
LGTM

PR #735, opened 2021-03-02
feat: add url redirect for stack report ui

deepak1725:
This PR adds API for URL Redirect to Stack Report UI
- Part of  [Add Report URL in CLI](https://issues.redhat.com/browse/APPAI-1702 )


bayesian/api/api_v2.py
```diff
@@ -307,6 +309,24 @@ def stack_analyses():
         raise HTTPError(500, e.args[0])
 
 
+@api_v2.route('/report/<stack_id>/<crda_key>', methods=['GET'])
```
arajkumar, line 312: Can't we extract crda_key from stack_id from the RDS?

prashbnair (changes requested):

bayesian/api/api_v2.py
deepak1725: Nope, we need a user reference even for RDS call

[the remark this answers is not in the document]
deepak1725: sure, do you want user_id or crda_key, actually in cli we are calling it as `crda_key`

arajkumar: why do you need user reference for rds access? just stack id is enough, thats how SA GET works.

arajkumar: every stack id associated with user id.

arajkumar: look at https://github.com/fabric8-analytics/fabric8-analytics-server/blob/44ab7d7ac248aeb2e14ae7f6fbc3a57e5e8e2ae6/bayesian/utils.py#L151.
it just needs request id.

deepak1725: I don't feel we need a DB call for this purpose at all. 
All DB calls add to monetary costs which in this case could be avoided.

arajkumar: The idea here is to make the url more concise. @prashbnair WDTY?

prashbnair: I am fine with the current URL.

prashbnair (approved):

bayesian/default_config.py
```diff
@@ -58,3 +58,8 @@
 COMPONENT_ANALYSES_LIMIT = int(os.getenv('COMPONENT_ANALYSES_LIMIT', 10))
 
 STACK_ANALYSIS_REQUEST_TIMEOUT = int(os.getenv('STACK_ANALYSIS_REQUEST_TIMEOUT', 120))
+
+THREESCALE_ACCOUNT_SECRET = os.getenv('THREESCALE_ACCOUNT_SECRET', "")
+BAYESIAN_API_HOSTNAME = os.getenv('BAYESIAN_API_HOSTNAME', 'http://recommendationapi-server:5000')
+STACK_REPORT_UI_HOSTNAME = os.getenv('STACK_REPORT_UI_HOSTNAME',
+                                     'https://stack-analytics-report.openshift.io')
```
arajkumar, line 65: Please add corresponding entries in template.yaml

bayesian/api/api_v2.py
```diff
@@ -303,6 +305,24 @@ def stack_analyses():
         raise HTTPError(500, e.args[0])
 
 
+@api_v2.route('/stack-report/<crda_key>/<stack_id>', methods=['GET'])
```
arajkumar, line 308: Is it okay to print and share the uuid?

bayesian/api/api_v2.py
```diff
@@ -303,6 +305,24 @@ def stack_analyses():
...
+        "access_token": "undefined",
+        "route_config": {
+            "api_url": BAYESIAN_API_HOSTNAME,
+            "ver": "v3",
+            "uuid": crda_key
+        },
+        "user_key": THREESCALE_ACCOUNT_SECRET
+    })
+    final_path = path + query_params
+    logger.info("Redirected to URL: %s ", final_path)
```
arajkumar, line 322: Can this be a debug print? we already have access logs enabled from gunicorn

arajkumar (changes requested):

bayesian/api/api_v2.py
deepak1725: No, I don't think so, This logs redirect URL, which I feel is not logged by Gunicorn

bayesian/default_config.py
deepak1725: Ack

bayesian/api/api_v2.py
arajkumar: I understand that, what I meant here is that we have minimal info print already in the form access logs for request.

bayesian/api/api_v2.py
arajkumar: I feel uuid is a sensitive value., shouldn't be shared as part of this.

arajkumar (changes requested):

openshift/template.yaml
```diff
@@ -185,6 +187,11 @@ parameters:
   required: false
   name: BAYESIAN_API_HOSTNAME
 
+- description: Stack Report UI (will be of Prod, if empty)
+  displayName: Stack Report UI
+  required: false
+  name: STACK_REPORT_UI_HOSTNAME
+
```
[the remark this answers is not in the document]
deepak1725: we are using default values in `default_config`, what's the difference..? have we adopted any new standards..?

bayesian/api/api_v2.py
deepak1725: We can mask it, base64 encoded or whatever

prashbnair: I think sharing UUID is not an issue. It is required when you are using CI so why should it be different here?

bayesian/api/api_v2.py
deepak1725: @arajkumar This is a really helpful log, I don't think we should skip it.. It logs which URL is built by combining a number of variables.

bayesian/default_config.py
```diff
@@ -58,3 +58,8 @@
 COMPONENT_ANALYSES_LIMIT = int(os.getenv('COMPONENT_ANALYSES_LIMIT', 10))
 
 STACK_ANALYSIS_REQUEST_TIMEOUT = int(os.getenv('STACK_ANALYSIS_REQUEST_TIMEOUT', 120))
+
+THREESCALE_ACCOUNT_SECRET = os.getenv('THREESCALE_ACCOUNT_SECRET', "")
```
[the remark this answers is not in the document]
deepak1725: @arajkumar That's what we are using for API authentication 
https://github.com/fabric8-analytics/fabric8-analytics-auth/blob/master/fabric8a_auth/auth.py#L101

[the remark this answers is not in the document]
arajkumar: @deepak1725 thats what my point is. Stack report UI needs 3scale user key, not the pre shared key(which is private) and used between 3scale gateway and api-server.

external --> 3scale-gateway --> api-server

external to 3scale-gateway: 3scale user key(e.g. prod: 9e7da76708fe374d8c10fa752e72989f, staging: 3e42fa66f65124e6b1266a23431e3d08)
3scale-gateway to api-server: PSK, private(not to be shared)

bayesian/api/api_v2.py
arajkumar: In CI, user/admin takes care copying it to CI specific secrets.

However when we display it over cli, user can share it with external party which is not desirable to know the UUID.

openshift/template.yaml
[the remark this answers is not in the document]
arajkumar: it is not new standard or anything, instead of having a text "Stack Report UI (will be of Prod, if empty)", express the default value over here itself.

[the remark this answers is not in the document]
deepak1725: Ack

bayesian/api/api_v2.py
deepak1725: We decided to remove UUID from URL, and handle it in API.

bayesian/api/api_v2.py
```diff
@@ -326,7 +326,7 @@ def stack_report_url(stack_id):
             "ver": "v3",
             "uuid": crda_key
         },
-        "user_key": THREESCALE_ACCOUNT_SECRET
+        "user_key": THREESCALE_ACCOUNT_VALUE
```
arajkumar, line 329: Use consistent naming. It should be called as  THREESCALE_USER_KEY

bayesian/api/api_v2.py
deepak1725: Update: RDS query to fetch crda_key corresponding to particular stack _id will be made in Redirect URL function only.

bayesian/api/api_v2.py
deepak1725: Ack

bayesian/api/api_v2.py
```diff
@@ -303,6 +305,34 @@ def stack_analyses():
         raise HTTPError(500, e.args[0])
 
 
+@api_v2.route('/stack-report/<stack_id>', methods=['GET'])
+def stack_report_url(stack_id: str):
```
arajkumar, line 309: Didn't we decided to do the RDS access on SA GET request when UUID is not present?

deepak1725: No, Prashant and I re-thought that and concluded on "RDS call in the Redirect URL" Approach would be a cleaner one.

prashbnair: The auth decorator expects a UUID , it is better to be consistent than to add hacks to work around that.

arajkumar: @prashbnair , the existing decorator can be kept as is. Without uuid, it just skips RDS call and sets the user type as unregistered. All we need to do is add one more check(the same one which @deepak1725 had put into the redir endpoint) in the SA GET which changes the user registration status after obtaining the UUID. It would be simple too and RDS access will be under 3scale which was raised as a concern during our last discussion.

deepak1725: @arajkumar Stack Report UI doesn't accept empty `uuid/crda_key`

prashbnair: @arajkumar It does not make for the user registration status to handled in different places, the auth filter does it.  That was the main reason for doing it this way. 
We need to ensure that there cannot be any SQL injection and we should be fine.

bayesian/api/api_v2.py
```diff
@@ -54,6 +54,8 @@
 from werkzeug.exceptions import BadRequest
 from f8a_utils.ingestion_utils import unknown_package_flow
 from f8a_utils import ingestion_utils
+from bayesian.default_config import THREESCALE_USER_KEY, \
+    BAYESIAN_API_HOSTNAME, STACK_REPORT_UI_HOSTNAME
```
arajkumar, line 58: nit
```suggestion
from bayesian.default_config import (THREESCALE_USER_KEY,
                                                                BAYESIAN_API_HOSTNAME, 
                                                                STACK_REPORT_UI_HOSTNAME)
```

arajkumar (approved):
lgtm, with a nit.
please wait for @prashbnair approval.

bayesian/api/api_v2.py
deepak1725: Ack

PR #736, opened 2021-03-05
fix: Implement concurrent batch data fetcher for component analysis

arajkumar:
Implements https://issues.redhat.com/browse/APPAI-1759

Implementation is based out of [ThreadPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#threadpoolexecutor). When it is run under gevent, Thread object will behave like a coroutine and it is not as heavy weight as normal OS threads. 

For each package batch, high-order function would be created and added into list. ThreadPoolExecutor::map function is applied on the high-order function list which runs it on the co-routine and result would be concatenated at the end.

#### config
```
batch-size: 10
concurrency-limit: 10
---
api-server-replica: 1
gremlin: 3
```

#### command
```
hey -m POST -D /tmp/req.json -H "content-type: application/json" -H "x-3scale-account-secret: not-set" -q 1 -z 30s -c 1 "http://bayesian-api-arajkumar-crda.apps.ocp.lab.psi.redhat.com/api/v2/component-analyses"
```

#### 10 packages:

> Note: For package count 10, since batch size is 10 both sequential and parallel should have approximately same numbers

Packages | Slowest(s) | Slowest(p) | Fastest(s) | Fastest(p) | Average(s) | Average(p) | Requests/sec(s) | Requests/sec(p)
-- | -- | -- | -- | -- | -- | -- | -- | --
10 | 2.2211 | 1.4891 | 0.5951 | 0.6051 | 0.8127 | 0.7526 | 0.9449 | 0.966
20 | 1.7681 | 1.7213 | 0.7963 | 0.6022 | 0.9695 | 0.7897 | 0.9641 | 0.9784
60 | 8.0089 | 2.039 | 1.5413 | 0.7038 | 2.197 | 0.8258 | 0.4408 | 0.944
> (s) - sequential
> (p) - parallel

Signed-off-by: Arunprasad Rajkumar <ar.arunprasad@gmail.com>


tests/utility/v2/test_component_analyses.py
```diff
@@ -283,10 +283,6 @@ def test_get_batch_ca_data_semver(self, _mockca):
         self.assertIsInstance(result, dict)
         self.assertIn('result', result)
         self.assertIsInstance(result.get('result'), dict)
-        self.assertIn('requestId', result)
```
arajkumar, line 286: These fields are not used in bossiness logic, and concurrent data fetcher ignores while doing concat

tests/utility/test_db_gateway.py
```diff
@@ -37,11 +37,11 @@ def setUpClass(cls):
         cls.ver = '1'
         cls.pkg = 'pkg'
         # Read Vendor Data from JSON.
-        gremlin_batch_data = os.path.join('/bayesian/tests/data/gremlin/gremlin_batch_data.json')
```
arajkumar, line 40: Remove absolute path from tests enable to execute tests locally without building docker images

bayesian/utility/v2/component_analyses.py
```diff
@@ -131,22 +144,28 @@ def get_batch_ca_data(ecosystem: str, packages) -> dict:
...
+        get_pseudo_data = functools.partial(GraphAnalyses.get_batch_ca_data_for_pseudo_version,
+                                            ecosystem)
+        graph_data_fetcher += list(_fetcher_in_batches(get_pseudo_data, pseudo_version_packages))
+
+    response = {
+        "result": {
+            "data": []
+        }
+    }
+    result = EXECUTOR.map(lambda f: f(), graph_data_fetcher)
```
dgpatelgit, line 162: Things looks good by review, but do ample testing with pseudo version + sem-version, only sem-version, only pseudo version use- cases.

bayesian/default_config.py
```diff
@@ -55,6 +55,4 @@
                      host=os.getenv('F8A_GEMINI_SERVER_SERVICE_HOST'),
                      port=os.getenv('F8A_GEMINI_SERVER_SERVICE_PORT'))
 
-COMPONENT_ANALYSES_LIMIT = int(os.getenv('COMPONENT_ANALYSES_LIMIT', 10))
-
 STACK_ANALYSIS_REQUEST_TIMEOUT = int(os.getenv('STACK_ANALYSIS_REQUEST_TIMEOUT', 120))
```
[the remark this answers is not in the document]
arajkumar: I just moved pieces those are relevant to this PR. It could be done in a separate PR.

bayesian/utility/v2/component_analyses.py
arajkumar: I guess the existing unit tests covers those scenarios.

dgpatelgit (approved):
lgtm

bayesian/utility/v2/component_analyses.py
```diff
@@ -112,7 +118,14 @@ def ca_validate_input(input_json: Dict, ecosystem: str) -> Tuple[List[Dict], Lis
     return packages_list, normalised_input_pkgs
 
 
-def get_batch_ca_data(ecosystem: str, packages) -> dict:
+def _fetcher_in_batches(func: Callable, packages: List,
+                        batch_size: int = COMPONENT_ANALYSES_SETTINGS.batch_size) -> Callable:
+    """Create partial function with batch package association."""
+    for i in range(0, len(packages), batch_size):
+        yield functools.partial(func, packages[i:i + batch_size])
```
prashbnair, line 125: This seems like an incorrect use of partial, you are actually not creating a partial function here but are completing the function definition.

Also is yield required here, the list is being generated completely so does it make sense to create a generator here?

prashbnair (changes requested):

bayesian/utility/v2/component_analyses.py
arajkumar: Here the partial function used to bind the args which are batched into callable. List of callable passed into `map` which executes these callable in the context of thread executor.

in this case use of yield is not for lazy eval, but just to avoid list creation in the function which generates batch.

bayesian/utility/db_gateway.py
```diff
@@ -100,7 +103,7 @@ def post_gremlin(cls, query: str, bindings: dict = None) -> dict:
         response = post(url=gremlin_url, data=json.dumps(payload))
         response.raise_for_status()
         elapsed_time = time.time() - started_at
-        logger.info("It took %s to fetch results from Gremlin.", elapsed_time)
+        logger.debug("gremlin call took %s sec", elapsed_time)
```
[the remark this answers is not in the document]
arajkumar: Okay

prashbnair (approved):

PR #737, opened 2021-03-08
[APPAI-1776] Fixed sentry error due to invalid version

dgpatelgit:
Fixes issue with version parsing during CA post call. There was version string clean up required for `maven` ecosystem, but this generic code was applied to all ecosystem. Also reduced log level from error to warning as CA response is served with return status 202 including other packages with valid version. Package with invalid version will be returned as `unknown_packages` in the response. 

Fixes jira issue :: https://issues.redhat.com/browse/APPAI-1776

Reduced vulnerabilities from `22` to `1` and resolved `1` exploits by upgrading dependent package versions in manifest file.

Address below sentry errors:
Stage :: https://sentry.stage.devshift.net/sentry/fabric8-analytics-stage/issues/202486/
Prod :: https://sentry.devshift.net/sentry/fabric8-analytics-production/issues/250686/


bayesian/utils.py
```diff
@@ -207,16 +207,19 @@ def convert_version_to_proper_semantic(version, package_name=None):
...
-        1.5.2 - RELEASE for semantic version to work."""
-        version = version.replace('.', '-', 3)
-        version = version.replace('-', '.', 2)
-        # Needed to add this so that -RELEASE is account as a Version.build
-        version = version.replace('-', '+', 3)
+
+        if ecosystem == "maven":
+            """Needed for maven version like 1.5.2.RELEASE to be converted to
+            1.5.2 - RELEASE for semantic version to work."""
+            version = version.replace('.', '-', 3)
```
yzainee, line 214: what value will be sent for ingestion in case the pkg is unknown? Also, is the version in graph going to be stored with . or - ? While fetching details from graph, will it expect in . or - form?

dgpatelgit: This functions is used to format the version we report in recommended version field of CA POST call. This is not used for any query graph / ingestion purpose. Also I don't know why `maven` version are converted from '.' to '-' forms, this generic conversion was creating issue for golang for sure, so moved maven specific code under ecosystem check.

yzainee: ok. can you ask the original coder to check why do we need this in the first place?

dgpatelgit: @deepak1725 Can you add oldie who can provide us reasoning behind this strange version transformation for Maven ecosystem?

dgpatelgit (commented):
@yzainee You suggest to check on version_comparator to remove this customize version transformation in API server. I did integrated version_comparator, but we cannot use this library directly because of following reasons.
1. Version comparator is used for version comparison only we cannot transform them to form a tuple with major, minor, patch and build. This tuple is refereed further in complete flow of CA response builder.
2. Invalid version string validation is missing in version comparator. This functionality is done by existing function in API server, which is a valid use-case.
3. With golang, we have to support semver and pseudo version format, which is taken care by existing function.

bayesian/utils.py
deepak1725: This code was written 3 years ago, You will hardly found the author of that.

yzainee:
@dgpatelgit
1-  why do you need major, minor patch etc? all you should care is about the version. how does it matter where it fits?
2- Please specify what type of validations are you referring to
3- FYI, im asking you to use version comparator just to find which is the latest etc. where else do you see a use of tuple and all?

dgpatelgit:
> @dgpatelgit
> 1- why do you need major, minor patch etc? all you should care is about the version. how does it matter where it fits?
> 2- Please specify what type of validations are you referring to
> 3- FYI, im asking you to use version comparator just to find which is the latest etc. where else do you see a use of tuple and all?

We need to validate version string for invalid version like 'upstream/1.0.0' or 'kubernetes.1.43.0' that we get in the request. Such validation are not provided by version comparator. I verified the current test cases using version comparator, not every thing can be handled. One such example is if we have  'upstream/1.0.1' (invalid version) and empty version (which is default HIGH version), current code return empty version as other version is invalid, but same code in version comparator will return 'upstream/1.0.1' as high version. This is not correct.

yzainee:
version comparator, as the name suggests, is used to compare versions. Its not a version validator. If there is any validation which is needed, it can be done in the code outside its purview. The comment is just to stop using tuple so that comparison can be done easily. how is tuple anyways helping you solve and check version;s validation?

dgpatelgit:
@yzainee Finally you wish came true, I have removed customer function and use version comparator to device highest non-cve version for recommendation.

bayesian/utility/v2/ca_response_builder.py
```diff
@@ -159,26 +159,25 @@ def get_version_without_cves(self, latest_non_cve_versions):
...
-                highest_version_tuple = version_info_tuple(
-                    convert_version_to_proper_semantic(highest_version)
-                )
-                # If version to recommend is closer to what a user is using then, use less than
-                # If recommendation is to show highest version then, use greater than
-                if graph_version_tuple > highest_version_tuple:
-                    highest_version = version
+        try:
+            input_version_comparable = ComparableVersion(self.version)
+            for version in latest_non_cve_versions:
```
yzainee, line 165: latest non cve version will never return a list/set. Its always a single value. Why are we using a loop here?

yzainee (changes requested):

bayesian/utility/v2/ca_response_builder.py
dgpatelgit: In the existing flow this is function variable and comes as a list. Below was the code that populates this parameter:
`latest_non_cve_versions = result_data[0].get('package', {}).get('latest_non_cve_version', [])`

yzainee: latest_non_cve_version = mgmt.makePropertyKey('latest_non_cve_version').dataType(String.class).make();
 schema for this value. it will not allow to store a list

dgpatelgit: We use gremlin to fetch the data, below is batch gremlin query we are using in CA POST flow.

ca_batch_query = """
            epv = [];packages.each {g.V().has('pecosystem', ecosystem).has('pname', it.name)
            .has('version', it.version).as('version', 'cve').select('version').in('has_version')
            .dedup().as('package').select('package', 'version', 'cve')
            .by(valueMap()).by(valueMap()).by(out('has_snyk_cve')
            .valueMap().fold()).fill(epv);};epv;
            """

Out of this query provides packages with non-cve version as an list, though it has always one value, but technically its a list.

dgpatelgit: In fact most of the properties are list as shown in snypet of response below:

"package": {
                   "ecosystem": [
                        "npm"
                    ],
                    "gh_subscribers_count": [
                        5
                    ],
                    "gh_contributors_count": [
                        -1
                    ],
                    "latest_version_last_updated": [
                        "20191117"
                    ],
                    "vertex_label": [
                        "Package"
                    ],
                    "libio_dependents_repos": [
                        "634788"
                    ],
                    "latest_non_cve_version": [
                        "7.0.3"
                    ],
                    "gh_issues_last_year_opened": [
                        -1
                    ],
                    "gh_issues_last_month_closed": [
                        -1
                    ],
                    "gh_open_issues_count": [
                        5
                    ],
                    "libio_dependents_projects": [
                        "290"
                    ],
                    "latest_version": [
                        "7.0.3"
                    ],
                    "tokens": [
                        "cliui"
                    ],
                    "package_relative_used": [
                        "not used"
                    ],
                    "gh_stargazers": [
                        210
                    ],
                    "gh_forks": [
                        16
                    ],
                    "package_dependents_count": [
                        -1
                    ],
                    "gh_prs_last_month_opened": [
                        -1
                    ],
                    "gh_issues_last_year_closed": [
                        -1
                    ],
                    "last_updated": [
                        1.6046478431252089E9
                    ],

yzainee: in that case you can pick [0]

dgpatelgit: @yzainee I would have not touched the existing logical flow, however as per above discussion I have removed latest non-cve into simple comparison check. PTAL

yzainee (approved):
LGTM

bayesian/utility/v2/ca_response_builder.py
```diff
@@ -20,7 +20,7 @@
 
 from bayesian.utility.db_gateway import GraphAnalyses
 from f8a_utils.user_token_utils import UserStatus
-from bayesian.utils import version_info_tuple, convert_version_to_proper_semantic
+from f8a_version_comparator.comparable_version import ComparableVersion
```
arajkumar, line 23: You need to explicitly add f8a_version_comparator in requirments.in.

bayesian/utility/v2/ca_response_builder.py
```diff
@@ -159,26 +159,21 @@ def get_version_without_cves(self, latest_non_cve_versions):
...
+        try:
+            input_version_comparable = ComparableVersion(self.version)
+
+            # latest non-cve is list with only one entry.
+            latest_non_cve_version_comparable = ComparableVersion(latest_non_cve_versions[0])
+
+            if latest_non_cve_version_comparable > input_version_comparable:
+                highest_version = latest_non_cve_versions[0]
+
+        except Exception as e:
```
arajkumar, line 172: Don't cache generic exception, Catch the specific one which will be thrown by ComparableVersion.

bayesian/utility/v2/ca_response_builder.py
```diff
@@ -159,26 +159,21 @@ def get_version_without_cves(self, latest_non_cve_versions):
...
+            logger.error(f"Package {self.package} @ {self.version} raised exception {e}")
```
arajkumar, line 173: It is not recommended to use preformatted string in logging. it would expanded regardless of log level which is not good. 
```suggestion
            logger.exception("Unable to parse version %s@%s", self.package, self.version)
```

bayesian/utility/v2/ca_response_builder.py
```diff
@@ -159,26 +159,21 @@ def get_version_without_cves(self, latest_non_cve_versions):
...
+
+        logger.info("Highest non-cve version for "
```
arajkumar, line 175: Ditto

bayesian/utility/v2/ca_response_builder.py
```diff
@@ -159,26 +159,21 @@ def get_version_without_cves(self, latest_non_cve_versions):
...
-                if graph_version_tuple > highest_version_tuple:
-                    highest_version = version
+        try:
+            input_version_comparable = ComparableVersion(self.version)
+
+            # latest non-cve is list with only one entry.
+            latest_non_cve_version_comparable = ComparableVersion(latest_non_cve_versions[0])
+
+            if latest_non_cve_version_comparable > input_version_comparable:
+                highest_version = latest_non_cve_versions[0]
```
arajkumar, line 170: This must go out of exception else block.

arajkumar (changes requested):

bayesian/utility/v2/ca_response_builder.py
dgpatelgit: This dependency was existing in requirements.txt, I am just using it here. Anyways it should be present in requirements.in also. Look like there was pending issue. I will fix it.

bayesian/utility/v2/ca_response_builder.py
dgpatelgit: This  variables are declared and initialized within this block, its good to perform this one step within the block?

arajkumar: you can still access those from the exception else block.

bayesian/utility/v2/ca_response_builder.py
arajkumar: requirements.txt contains both direct and trans dependencies., what if we stop using fa8_utils in future?

yzainee: @arajkumar in that case, the unit test will fail and will get to know about a missing dep

bayesian/utility/v2/ca_response_builder.py
dgpatelgit: Yes, your right. From interface perspective, compare_to() interface of `ComparableVersion` can throw TypeError(), So as a caller it better we handle that exception. Also I don't see a big advantage of moving this condition statement outside the try catch block.

bayesian/utility/v2/ca_response_builder.py
arajkumar: We are following this as a convention across the repo. requirement.in must be a subset of requirements.txt

bayesian/utility/v2/ca_response_builder.py
arajkumar: where is that described?

dgpatelgit: If you go by documentation, even `ComparableVersion`  should also not throw any error as its not documented, so we don't need this try catch block itself. Let me move this out, just to make you happy.... I still feel its better to keep this under a try catch and does not add any overhead.

bayesian/utility/v2/ca_response_builder.py
```diff
@@ -159,26 +160,20 @@ def get_version_without_cves(self, latest_non_cve_versions):
...
+            # latest non-cve is list with only one entry.
+            latest_non_cve_version_comparable = ComparableVersion(latest_non_cve_versions[0])
+        except TypeError:
+            logger.error("Package %s@%s raised a TypeError", self.package, self.version)
+
+        if latest_non_cve_version_comparable > input_version_comparable:
+            highest_version = latest_non_cve_versions[0]
+
+        logger.info("Highest non-cve version for %s@%s is %s", self.package, self.version,
+                    highest_version)
```
arajkumar, line 176: I've stated that the comparison must go into exception else block, this is incorrect.
```suggestion
        try:
            input_version_comparable = ComparableVersion(self.version)

            # latest non-cve is list with only one entry.
            latest_non_cve_version_comparable = ComparableVersion(latest_non_cve_versions[0])
        except TypeError:
            logger.error("Package %s@%s raised a TypeError", self.package, self.version)
        else:
            if latest_non_cve_version_comparable > input_version_comparable:
                highest_version = latest_non_cve_versions[0]

        logger.info("Highest non-cve version for %s@%s is %s", self.package, self.version,
                    highest_version)
```

tests/test_utils.py
```diff
@@ -212,49 +212,31 @@ def test_github_url_exist_or_not(self):
...
+    @patch('bayesian.utils.init_celery', return_value=None)
+    @patch('bayesian.utils.run_flow', return_value=1234)
+    def test_server_create_analysis(self, _rf_mock, _cf_mock):
+        """Verify various combinations of create analysis function."""
+        assert server_create_analysis('pypi', 'pkg1', '1.5.4', None, False, False, False) == 1234
+
+        assert server_create_analysis('pypi', 'pkg1', '1.5.4', None, True, False, False) == 1234
+
+        pkg = 'debug:artifact:0.4.3'
+        assert server_create_analysis('maven', pkg, '0.4.3', None, False, False, False) == 1234
```
arajkumar, line 239: What is the purpose of these tests? why does it only asserts mocked func return values which will be always same?

arajkumar (changes requested):

tests/test_utils.py
dgpatelgit: Reason was Codecov, after removing existing util functions codecov was failing due to reduction in coverage percentage (it was as low as 18% for utility file). You can see few commit before error on Codecov. To improve unit test coverage I added some of missing test cases for run-flow utility functions.

arajkumar: That is fine as the number of LOCs are reduced. But these tests doesn't make any sense to me.

dgpatelgit: @arajkumar what do you suggest? shall we reduce Codecov level from B to C. Currently we fail Codecov if any file has level less than 'B' which is 19% coverage.

arajkumar (approved):
lgtm

PR #738, opened 2021-03-12
chore: change user not found log level from exception to info

prashbnair:


arajkumar (commented):
@prashbnair Should we audit the use of all `log.exception` in api-server and replace it?

prashbnair:
> Should we audit the use of all log.exception in api-server and replace it?

I think there are cases where it should be used, lets take it on a case by case basis. I am not aware of any other log that comes up this often and is not useful.

arajkumar (approved):
lgtm

PR #739, opened 2021-03-18
fix: Use Jobs api for book-keeping task.

rafiu007:


bayesian/utils.py
```diff
@@ -19,9 +19,16 @@
 from f8a_worker.setup_celery import init_celery
 from .default_config import STACK_ANALYSIS_REQUEST_TIMEOUT
 from sqlalchemy.exc import SQLAlchemyError
+from f8a_utils.ingestion_utils import trigger_workerflow
 
 logger = logging.getLogger(__name__)
 
+_INGESTION_API_URL = "http://{host}:{port}/{endpoint}".format(
```
jparsai, line 26: You don't need this, its already defined in f8a_utils

jparsai (changes requested):

bayesian/utils.py
rafiu007: My bad ,forgot to remove this from last implementation.

bayesian/utils.py
```diff
@@ -65,20 +66,22 @@ def get_user_email(user_profile):
...
+            "packages_list": packages_list,
+            "user_id": headers.get('uuid', None),
+            "user_agent": headers.get('User-Agent', None),
+            "source": request_args.get('utm_source', None),
+            "telemetry_id": headers.get('X-Telemetry-Id', None)
         }
     }
-    return server_run_flow('componentApiFlow', args)
+
+    trigger_workerflow(payload)
```
jparsai, line 84: This can throw exception, so better to use try except to avoid any break in flow.

jparsai (changes requested):

bayesian/utils.py
rafiu007: If any exception is thrown wont it be inside jobs api or utils.where this has been handeled with try except.

jparsai (approved):

PR #772, opened 2021-07-06
chore: User caching.

jparsai:
Created new API to create user cache in PVC. API will be called from daily reports to update the cache.


yzainee (approved):
LGTM

yzainee (changes requested):
In template you need to add the secrets

yzainee:
example:
- name: SNYK_TOKEN
                  valueFrom:
                    secretKeyRef:
                      name: snyk-secrets
                      key: token

jparsai:
> In template you need to add the secrets
I am using an existing secret.
https://github.com/fabric8-analytics/fabric8-analytics-server/blob/a1e2e0a2eaaf94fc2055251c9ec36610503f527b/openshift/template.yaml#L129

yzainee (approved):

PR #775, opened 2021-07-13
chore: Use API to get user details 

jparsai:
# Changes
- Deleted the code which is not required anymore.
- Calling a Jobs API to get user details from cache if flag is enabled
- Calling the Jobs API to create a new entry in cache if a user submitted Snyk token

### DIFF
--- a/bayesian/utility/user_utils.py
+++ b/bayesian/utility/user_utils.py
@@ -10,10 +10,14 @@ from sqlalchemy.dialects.postgresql import insert
 
 from f8a_worker.models import (UserDetails)
 from f8a_utils.user_token_utils import UserStatus
-from bayesian.settings import ENABLE_USER_CACHING, DB_CACHE_DIR
+from bayesian.settings import ENABLE_USER_CACHING
 from bayesian import rdb
-import json
 import os
+import requests
+
+_JOB_API_URL = "http://{host}:{port}/internal/ingestions".format(
+    host=os.environ.get("JOB_SERVICE_HOST", "bayesian-jobs"),
+    port=os.environ.get("JOB_SERVICE_PORT", "34000"),)
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


bayesian/utility/user_utils.py
```diff
@@ -57,61 +61,28 @@ def create_or_update_user(user_id, snyk_api_token, user_source):
...
-                       "status": UserStatus.REGISTERED.name,
-                       "created_date": datetime.datetime.now(),
-                       "registered_date": datetime.datetime.now()}
+                       "status": UserStatus.REGISTERED.name}
         try:
-            with open(db_cache_file_path, 'w', encoding='utf-8') as file:
-                json.dump(user_detail, file, ensure_ascii=False, indent=4, default=str)
-
-            logger.info("User added in cache with id %s", user_id)
+            url = _JOB_API_URL + "/create_or_update_user_in_cache"
```
yzainee, line 71: check to see if caching is enabled should be done in api server itself. it will save us the api call

yzainee (changes requested):

yzainee:
if caching is disabled, where will it fetch the user data from?

bayesian/utility/user_utils.py
jparsai: Yes in L66 you can see if ENABLE_USER_CACHING is False it wont go inside and wont call jobs API.

jparsai:
> if caching is disabled, where will it fetch the user data from?

It will follow old approach to get user detaisl i.e. from RDS.
In auth function which is called before API we are validating user, here only we are switching between RDS/Cache based on env variable.
https://github.com/jparsai/fabric8-analytics-server/blob/ccf420d54bce689d721024156dfcd131a044c51a/bayesian/auth.py#L39

yzainee (approved):

PR #777, opened 2021-07-15
fix: Modify secret "3scale premium key".

rafiu007:
secret has to be added via app-sre. This has cused deployment failure and e2e failure.


sawood14012 (approved):
LGTM

yzainee-zz (approved):
LGTM. merge after secret is created in vault by app-sre for stage

PR #778, opened 2021-07-15
chore: add test cases

mathur07:


tests/api/test_api_v2.py
```diff
@@ -124,6 +124,86 @@ def test_get_component_analyses_bad_request(self):
...
+        accept_json = [('Content-Type', 'application/json;')]
+        response = self.client.post(
+            api_route_for('/vulnerability-analysis'), data=json.dumps(payload), headers=accept_json)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.json, self.recommendation_data)
+
+    def test_get_vulnerability_analysis_bad_request(self):
+        """VA POST: Bad Request."""
+        payload = {
+            "ecosys": 'pypi',
```
rafiu007, line 173: This should be "ecosystem" .
and For python this should not be the response.
currently only golang is not supported

rafiu007 (approved):
LGTM. Suggested a minior change.
Wait for review from @yzainee

yzainee-zz (approved):
LGTM. I would like to see a thing added though. As the fixed_in field is a list, try adding 2-3 values in the graph mock response and see if you get all the 3 in the response you send back from the api.

rafiu007 (approved):
LGTM

PR #790, opened 2021-08-09
fix: updated `get-token` spec.

rafiu007:


jparsai (approved):

api_specs/v2.2/component_analyses_v2.2.yaml
```diff
@@ -162,6 +162,7 @@ components:
   examples:
     higher-rate-3scale:
       value:
+        - url: "https://f8a-analytics-2445582058137.production.gw.apicast.io/api/v2/"
```
yzainee, line 165: are you not adding the source here too?

yzainee (changes requested):

api_specs/v2.2/component_analyses_v2.2.yaml
jparsai: It is an example, actual value will be returned from code only.

rafiu007: @yzainee I am not sure what would be the use of keeping the source field.

yzainee (approved):

PR #791, opened 2021-08-09
fix: update `get-token` api to return url as well.

rafiu007:


bayesian/api/api_v2.py
```diff
@@ -68,7 +68,8 @@ def get_token():
     try:
         # return default key
         THREESCALE_PREMIUM_USER_KEY = os.getenv('THREESCALE_PREMIUM_USER_KEY')
-        return jsonify({"key": THREESCALE_PREMIUM_USER_KEY}), 200
+        PREMIUM_URL = os.getenv('PREMIUM_URL')
+        return jsonify({"url": PREMIUM_URL, "key": THREESCALE_PREMIUM_USER_KEY}), 200
```
yzainee, line 72: corresponding changes in the unit tests should be done as well.

yzainee (changes requested):

bayesian/api/api_v2.py
```diff
@@ -68,7 +68,8 @@ def get_token():
     try:
         # return default key
         THREESCALE_PREMIUM_USER_KEY = os.getenv('THREESCALE_PREMIUM_USER_KEY')
-        return jsonify({"key": THREESCALE_PREMIUM_USER_KEY}), 200
+        PREMIUM_URL = os.getenv('PREMIUM_URL')
```
yzainee, line 71: PREMIUM_URL needs to be added to the template file

yzainee (changes requested):

bayesian/api/api_v2.py
rafiu007: Yes.On it.

bayesian/api/api_v2.py
rafiu007: IT is added @yzainee

rafiu007: which template are you talking about?

yzainee: openshift/template.yaml

yzainee: oh i didnt notice.

openshift/template.yaml
```diff
@@ -127,6 +127,11 @@ objects:
               secretKeyRef:
                 name: 3scale
                 key: three_scale_premium_user_key
+          - name: PREMIUM_URL
+            valueFrom:
+              secretKeyRef:
```
yzainee, line 132: if you are introducing a new secret, you will need this secret in the vault before you merge this PR

yzainee: I dont think this needs to be a secret. It can be a configurable value in app interface.

rafiu007: Since the token is a secret i made this also a secret.

bayesian/api/api_v2.py
rafiu007: I feel there is no modicfication required in the tests. The response code test is already done.

yzainee: add the new key "url" in the test

openshift/template.yaml
yzainee: there is no need to have this as a secret. please change it

rafiu007: Done.

yzainee (approved):

bayesian/api/api_v2.py
rafiu007: Done

PR #796, opened 2021-08-25
[Revert after rds upgrade]chore: Disable user registrations.

rafiu007:


bayesian/api/user_api.py
```diff
@@ -49,6 +49,11 @@ def create_or_update_user():
 
     user_id = content.get('user_id')
 
+    # Needs to be reverted after rds upgrade
+    if 1 == 1:
```
jparsai, line 53: why 1 == 1?

rafiu007: I didnt wanna remove other portion of the code just for this.
so instead of returning i thought of adding a condition whihc is always true.

rafiu007: Open to better methodologies.

jparsai (approved):

yzainee (approved):
LGTM

PR #797, opened 2021-08-25
[Revert after rds upgrade]chore: Disable SA call for rds upgrade.

rafiu007:


bayesian/api/api_v2.py
```diff
@@ -248,6 +254,10 @@ def stack_analyses():
     if request.method == 'GET':
         raise HTTPError(400, error="Request id missing")
 
+    # Revert after RDS upgrade.
```
jparsai, line 257: Why are you returning error in SA Get?

bayesian/api/api_v2.py
```diff
@@ -211,6 +211,12 @@ def stack_analyses_with_request_id(external_request_id):
     """Handle stack analyses report fetch api."""
     start = time.time()
     logger.debug("[GET] /stack-analyses/%s", external_request_id)
+    # Revert after RDS upgrade.
+    if True == True:
```
jparsai, line 215: It will misleading, it should be empty instead of dummy data, even if user runs for other ecosystems he/she will get python data in response.

jparsai (changes requested):

bayesian/api/api_v2.py
rafiu007: this was present from before.

bayesian/api/api_v2.py
rafiu007: yes.But @yzainee suggestion was to keep dumy data.
I am okay with naything.Either way UTs will fail.

jparsai: Yes, Yusuf suggested to have a dummy data, which will have empty for all fields, so clients will not fail because of missing keys but also show 0/empty for all analysis

bayesian/api/api_v2.py
jparsai: @mathur07 isn't it supposed to have uuid to show banner? or even it fails it will redirect to banner?

bayesian/api/api_v2.py
rafiu007: Ahh okay.

yzainee: Even this custom data should be ok. WDYT @jparsai ?

rafiu007: fixed it

yzainee (approved):

bayesian/api/api_v2.py
jparsai: @yzainee does it look fine to you? I am not sure if this is what we need.

yzainee: @rafiu007 I think we need to send a static id here.. right? @mathur07 can confirm

rafiu007: I can return some dummy id.Or In UI we can for sometim hardcode a dummy id

jparsai (approved):

bayesian/api/api_v2.py
rafiu007: @mathur07 @yzainee ^^

sawood14012 (approved):

.github/workflows/ci-build.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         scripts:
-          - run-linter
+        #  - run-linter
```
sawood14012, line 15: revert this line @rafiu007

sawood14012 (approved):

.github/workflows/ci-build.yml
rafiu007: done

PR #822, opened 2021-12-17
fix 'fixed_in' issue in VA

AkshayBhansali18:
fix 'fixed_in' issue in VA


jparsai (changes requested):
You can do it in one line as well 
`vulnerability.get("fixed_in", {})`

jparsai (approved):

